PR #18, opened 2026-02-24
Add decomposed reward shaping for exploration

beyarkay:
## Summary

- Adds three independent reward-shaping components that compare the current world against the solved factory (`_solved_world_CWH`), giving partial credit for each dimension separately:
  - `tile_match_location`: any entity placed where solution needs one (~1/64 random hit rate)
  - `tile_match_entity`: entity type matches solution at each tile (~1/2 random hit rate)
  - `tile_match_direction`: direction matches solution at each tile (~1/5 random hit rate)
- The agent never observes the solution; only the reward function uses this privileged info during training
- Updates `sweep.yaml` to search over the three shaping coefficients plus an expanded `ent_coef` range (0.001–0.05)
- Adds seed-consistency tests verifying `generate_lesson` determinism (critical since we call it twice with the same seed)

## Motivation

The agent struggles to explore because the probability of randomly getting the correct tile + entity + direction is ~1/640 per step. Previous experiments (PR #16) confirmed the bottleneck is exploration, not representational capacity. Rather than rewarding only fully-correct tiles (nearly as sparse as throughput), this decomposes the signal so each component is independently discoverable.

## Test plan

- [x] 21 new tests pass (6 reward shaping + 15 seed consistency)
- [x] All 211 existing tests pass
- [x] Ruff clean
- [x] PPO smoke test passes (5K timesteps)
- [ ] Run W&B sweep to find optimal shaping coefficients (add `sweep` label)


beyarkay:
It seems like this is significant iff you run paired t-tests:

>  PR #18 (reward-shaping): already significant! p=0.040
>
>  The pairing removes between-seed variance (std of diffs = 0.051 vs unpaired pooled sd = 0.123). Cohen's d jumps from 0.55 to 1.34. You'd only need 7 seeds for 80% power — you're already at 61.6% with 5.

I'll likely merge this, but will wait for more test results

PR #133, opened 2026-06-12
Overfit experiment: SFT on MOVE_ONE_ITEM only

beyarkay:
## What

Pin SFT dataset generation to **`MOVE_ONE_ITEM` only** so we can test — in the Karpathy sense — that the SFT pipeline can actually drive **`val/MOVE_ONE_ITEM/throughput`** up given ample compute. If throughput won't rise on the single simplest lesson, the problem is upstream of the lesson mix, not in the data diversity.

This is an **experiment scaffold, not destined for `main` as-is.** One-line revert: restore `kinds = list(LessonKind)` in `generate_dataset`.

## Changes

- `sft.py`: `generate_dataset` samples only `MOVE_ONE_ITEM` (other kinds commented out, not deleted).
- `tests/test_sft.py`: skip the two tests that assert the dataset spans ≥2 `LessonKind`s (they don't hold while pinned to one kind). Reason string points back here.

## Why this is a fair test (not memorize-3-layouts)

`build_factory` defaults to `random_item=True` and `max_entities=inf`, so each `MOVE_ONE_ITEM` lesson randomizes source/sink position, both directions, the item (full item set), and the path shape. Held-out val factories are genuinely unseen configs — so a rising `val/MOVE_ONE_ITEM/throughput` means real generalization on this lesson.

## Running the tracked RunPod job

RunPod is **label-gated** (not auto on PR open). Add the **`sft-smoketest`** label to trigger `scripts/ci/sft_smoke_test.sh`, which runs one tracked job:
`sft.py --size 11 --num-samples 300000 --epochs 30 --track` (logs to W&B `factorion`, 4h watchdog).

### DIFF
--- a/.github/workflows/ppo-train.yml
+++ b/.github/workflows/ppo-train.yml
@@ -47,6 +47,16 @@ on:
         required: false
         default: "0.02"
         type: string
+      lr_anneal_timesteps:
+        description: "Anneal LR -> min_lr over the first N env steps, then hold (decoupled from total_timesteps). 0 = legacy anneal-to-0 over the whole run"
+        required: false
+        default: "500000"
+        type: string
+      min_lr:
+        description: "LR floor held after lr_anneal_timesteps (keeps a long run improving instead of freezing at 0)"
+        required: false
+        default: "1e-5"
+        type: string
       seed:
         description: "Random seed"
         required: false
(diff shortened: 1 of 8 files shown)


beyarkay:
## ⚠️ TEMPORARY — revert the canonical-path change before this is real

Commit `702f97a` ("MOVE_ONE_ITEM: pick one canonical path") replaces
`random.choice(paths)` in `build_factory` with `min(paths, key=_canonical_path_key)`,
collapsing every (source, sink) to a **single** deterministic belt path.

**This is a derisk experiment ONLY** — to test whether multiple-shortest-path
ambiguity is what caps `val/MOVE_ONE_ITEM/dir_acc` (~0.78, with train==val loss
and only the direction head stuck). We explicitly **want** the generator to
produce lots of diverse belt layouts; forcing one canonical path throws that
diversity away and must not ship.

### TODO before merge / real training
- [ ] Revert `702f97a`: restore `chosen_path = random.choice(paths)` (and drop
      `_MOVE_PRIORITY` / `_canonical_path_key` from `factorion.py`).
- [ ] If the experiment confirms ambiguity is the cause, fix it the
      **diversity-preserving** way instead: a multi-target direction loss that
      rewards *any* valid frontier direction (computed from `find_belt_paths`),
      rather than supervising toward one arbitrary path.

Verdict pending run: https://github.com/beyarkay/factorion/actions/runs/27431599041

PR #140, opened 2026-06-15
SFT sweep plumbing: dynamic CNN architecture (--layers / --kernel-size)

beyarkay:
## What

Makes the CNN encoder architecture **sweepable** so a W&B Bayesian sweep can search depth, per-layer width, and receptive field against the greedy-build metric `val/throughput` (the default since #139).

Replaces the fixed `chan1/chan2/chan3/flat_dim` knobs with:

- **`layer1..layer8`** — each is a conv layer's channel width; `0` drops that layer (slots compacted in order). Depth *and* width become **independent ordinal** sweep dimensions (`int_uniform`), instead of one opaque categorical `"64,64,64"` string the optimiser can't reason about.
- **`kernel_size`** — receptive-field knob. `RF = 1 + n_layers·(kernel_size−1)`. Padding is pinned to `kernel_size//2` ("same"), because the per-tile heads require the feature map to stay exactly grid-sized (one logit per cell). Padding is therefore *not* a free knob.

`AgentCNN` itself takes a clean `layers` sequence + `kernel_size`; the dataclass slots compact via `layers_from_args()`.

## Sweep config

`sft_sweep.yaml` now:
- targets `metric: val/throughput` (was `val/acc`),
- searches `layer1..8` + `kernel_size{3,5}` + `lr` (lr co-swept since it covaries with depth),
- **pins** size 11 / 300k pairs / 50 epochs to the **C7 high-throughput regime** (held-out greedy 0.947, deep 6-layer RF-13 encoder),
- leaves per-head loss weights / schedule / regularisation at their `SFTArgs` defaults (the prior sweep's winner) to keep the search within `run_cap`.

The CI harness already supports SFT sweeps end-to-end (jobs `sft-sweep-setup` → `sft-sweep` → `sft-sweep-report`, triggered by the `sft-sweep` label); only the param names needed updating.

## Also

- `scripts/factory_builder.py` infers encoder arch (any depth/kernel) from the checkpoint instead of hardcoding `encoder.0/2/4`.
- `factorion-mod/server/server.py` translates its sidecar chans to `layers=`.
- PPO `sweep.yaml` + both `apply_*_sweep_best.py` updated for the renamed params (PPO shares `AgentCNN`).

## Verification

- `ruff check .` ✓
- full `pytest` — 3395 passed, 77 skipped ✓ (added depth/kernel artifact-name coverage; verified compaction, odd-kernel guard, all-zero guard)
- SFT smoke (`--layer1/2/3`) ✓ · PPO smoke ✓ — both build the encoder dynamically

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -821,7 +821,7 @@ jobs:
           SWEEP_URL="${{ steps.create.outputs.sweep_url }}"
           SWEEP_PODS="${{ inputs.sweep_agents || '1' }}"
           AGENTS_PER_POD="${{ inputs.sweep_agents_per_pod || '5' }}"
-          SWEEP_COUNT="${{ inputs.sweep_count || '20' }}"
+          SWEEP_COUNT="${{ inputs.sweep_count || '30' }}"
           TOTAL_AGENTS=$(( SWEEP_PODS * AGENTS_PER_POD ))
           gh pr comment "${{ github.event.pull_request.number }}" --body "$(cat <<EOF
           **SFT W&B Sweep started** — ${SWEEP_COUNT} iterations across ${SWEEP_PODS} pod(s) x ${AGENTS_PER_POD} agents/pod (${TOTAL_AGENTS} total agents)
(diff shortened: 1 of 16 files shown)


beyarkay:
Superseded — this mega-branch was split into focused PRs that have all landed: dynamic CNN encoder (#146), read_metric fix (#144), CI sweep count (#145), SFT timing (#147), dropout (#149), eval-from-empty (#148), per-kind eot (#153), direction diagnostics (#155/#156). The sweep it was plumbing already ran (ui3v0gn8 → run kkcv6xe3), and its winning hyperparameters are being adopted as defaults in #157. The experimental sweep YAML isn't being kept. Closing as out of date.

PR #143, opened 2026-06-16
SFT: speed up train/val loop + add timing logs

beyarkay:
## What

Speeds up SFT training and adds timing instrumentation. No behaviour change —
metrics are numerically identical to before.

The per-epoch train+val loop dominated wall time (~2 min/epoch for a tiny CNN
on GPU). The cause was per-batch `.item()` / `.tolist()` calls (~10 in train,
~70 in val once the per-kind loop is counted) — each forces a GPU→CPU sync
that stalls the pipeline, so the GPU never runs ahead.

## Changes

- **Defer syncs**: accumulate epoch sums in on-device tensors and sync **once**
  per loop (train + val) instead of every batch.
- **Vectorise per-kind val aggregation** with `index_add_`, replacing a
  per-batch python loop over kinds and its `.tolist()`/`.item()` syncs.
- **`pin_memory` + `non_blocking` transfers on CUDA only.** A `non_blocking`
  copy from *unpinned* memory races (may not finish before use → garbage tile
  index → out-of-bounds gather), so both are gated on `pin = cuda.is_available()`.
- **Timing logs** to track speedups run-to-run:
  - per-epoch: `| t: train=Xs val=Ys roll=Zs`
  - end-of-run: `Timing: total=… | train=… val=… rollout=… | train throughput=N samples/s`
  - W&B `perf/train_seconds`, `perf/val_seconds`, and summary-JSON fields.

## Validation

- **Numerically identical** to the old loop: byte-identical metrics on a fixed
  seed, and the `index_add_` per-kind aggregation was cross-checked in-process
  against the reference loop (passed on data with nonzero per-kind counts).
- Full test suite: **3413 passed, 77 skipped**. `ruff` clean.
- GPU speedup not benchmarked locally (dev box is MPS, where pin/non_blocking
  are intentional no-ops) — the new timing logs quantify it on the next CUDA run.

### DIFF
--- a/sft.py
+++ b/sft.py
@@ -714,8 +714,17 @@ def train_sft(args: SFTArgs):
         eot_labels[val_idx],
         lesson_kinds[val_idx],
     )
-    train_loader = DataLoader(train_ds, batch_size=args.batch_size, shuffle=True)
-    val_loader = DataLoader(val_ds, batch_size=args.batch_size)
+    # `pin` gates BOTH pin_memory and the non_blocking transfers below: an
+    # async (non_blocking) host->device copy is only safe from PINNED memory.
+    # With an unpinned source (CPU/MPS, where pin is False) a non_blocking copy
+    # races — it may not finish before the tensor is read, yielding garbage
+    # (e.g. a wild tile index -> out-of-bounds gather). So pinning + async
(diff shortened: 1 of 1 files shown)


beyarkay:
THis is just much slower ¯\_(ツ)_/¯

PR #154, opened 2026-06-17
fix(sft): W&B artifact metadata read removed chan1/2/3 attrs

beyarkay:
## Problem

A tracked SFT run crashes **after** a full training run completes, while logging the W&B artifact:

```
Checkpoint saved to: sft_checkpoint.pt
Summary written to /workspace/factorion/sft_summary.json
Traceback (most recent call last):
  File "sft.py", line 1432, in <module>
    train_sft(args)
  File "sft.py", line 1432, in train_sft
    "chan1": args.chan1,
AttributeError: 'SFTArgs' object has no attribute 'chan1'
```

#146 renamed the encoder hyperparams `chan1/2/3` → `layer1..8` on `SFTArgs`, but the artifact-metadata dict in `train_sft` kept reading `args.chan1/2/3`.

## Why it slipped through CI

That metadata dict lives inside `if args.track and run is not None:`. Every existing end-to-end test runs with the `track=False` default, so the entire tracking branch was never exercised — the crash only appears on a real tracked (GPU) run, and only at the very end after training + checkpointing have already burned the compute.

## Fix

Record the resolved encoder widths via `layers_from_args(args)` (a single `"layers"` key) instead of the three removed `chan*` attributes. The artifact *name* already encodes layer depth/width via `_artifact_name`, so this just keeps the metadata consistent.

## Test

Adds `TestTrainSFTTrackingArtifact::test_tracking_branch_runs_end_to_end`, which drives `train_sft(track=True)` against a **fake `wandb`** injected into `sys.modules` (no network, no creds, no wandb install needed — runs identically on CI). This exercises the whole tracking branch end-to-end: `init` → per-epoch `run.log` → summary writes → artifact metadata construction → `add_file` → `log_artifact` → `finish`. It is a verified red→green regression test: it reproduces the exact `AttributeError` on the old code and passes with the fix.

## Checklist (Python-only diff; no Rust touched)

- ✅ `ruff check .`
- ✅ Full Python test suite — 3417 passed, 77 skipped
- ✅ PPO smoke test
- ✅ SFT smoke test


beyarkay:
duplicate of #152

PR #158, opened 2026-06-17
factory UI: default to the canonical kkcv6xe3 checkpoint

beyarkay:
## What

The factory-builder UI (`scripts/factory_builder.py`) auto-loads a W&B checkpoint at launch when neither `--checkpoint` nor `--wandb-run` is given. Point that default (`Args.wandb_run`) at run **kkcv6xe3** (sft-11x11, best_val_throughput 0.335) — the SFT checkpoint we're standardising on — instead of the older `xlnqmtzt`. The UI resolves the run id to its newest `type=model` artifact, which kkcv6xe3 has (`sft-s11-n1m-e45-bs512-lr3.2e-3-c93-69-96:v0`).

Pinned with `TestDefaultCheckpoint` so the canonical default can't drift silently.

Companion to #157 (which makes the *training* defaults reproduce kkcv6xe3). Independent — separate file, can merge in any order.

### DIFF
--- a/scripts/factory_builder.py
+++ b/scripts/factory_builder.py
@@ -111,10 +111,11 @@ class Args:
     checkpoint: Optional[str] = None
     """path to a trained SFT/PPO checkpoint (.pt). If set, the UI shows
     the model's predicted next placement and exposes an Apply button."""
-    wandb_run: Optional[str] = "xlnqmtzt"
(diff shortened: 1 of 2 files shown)


beyarkay:
Folded into #157 (the canonical-kkcv6xe3 PR) — the `wandb_run` default change now lives there, without the magic-number pin test. Closing this in favour of the combined PR.

PR #170, opened 2026-06-22
Textual factory format for unit tests

beyarkay:
## What

A compact, human-writable **textual factory format** for the Rust resolver, so factory unit tests can be described as text, parsed into a `World`, and asserted (throughput and/or graph). Lives in `factorion_rs/src/textual.rs` as a `#[cfg(test)]` module — it never enters the shipped cdylib, and `serde`/`serde_yaml` are dev-dependencies only.

## The format

A fixture is **pure YAML**. The grid is a `factory:` block scalar; sibling keys carry the assertions:

```yaml
description: |
  A source feeds two belts into a sink.
items:
- { x: 0, y: 0, item: copper_cable }   # source emits
- { x: 3, y: 0, item: copper_cable }   # sink counts
throughput:
- { item: copper_cable, per_second: 15 }
graph: |
  S@0,0 -> b@1,0
  b@1,0 -> b@2,0
  b@2,0 -> K@3,0
factory: |
  S> b> b> K> ..
```

- **Grid encoding** — each tile is 2 chars + 1 filler (*gobble 2, skip 1*). Direction is any of `^>v<`; entity is any registry char (`b` belt, `i` inserter, `a` assembler 3×3, `Y` splitter 2×1, `d`/`u` underground down/up, `S` source, `K` sink). `..` empty, two spaces = multi-tile interior, `#` full-line comments. The grid is strict: unexpected characters (e.g. `b@`) and non-rectangular rows are errors.
- **Multi-tile** entities are drawn across their full footprint (assembler = hollow box, splitter = `YvvvY` bracket / stacked `Y>`), generic over `Item::size()` — a new entity is one row in `ENTITY_CHARS`.
- **`items:`** binds an item to the entity at `(x,y)` (multi-tile by anchor): source emit / sink count / assembler recipe.
- **`throughput:`** and **`graph:`** are each optional (at least one required). `graph:` is `A@x,y -> B@x,y` per line, compared as a set against `build_graph`'s edges; an empty graph asserts *no* edges.
- Several factories per file via YAML's native `---` document separator (`parse_many`).

## API

`crate::textual::{parse, parse_many, assert_throughput, check_throughput, assert_graph, check_graph, render}`. `render` is the inverse of the grid parser (geometry round-trips).

## Tests

- 30+ unit tests: tokenizer, multi-tile placement, YAML header, item bindings, render round-trip, malformed-input rejection (junk chars, ragged rows, bad graph specs), and ports of existing `throughput.rs` cases.
- A directory sweep (`tests/factories/*.yaml`) parses every fixture and asserts its declared `throughput:`/`graph:`, with rich failure output (file path, factory index, description, expected-vs-actual). Fixtures include a graph-only catalogue of two-entity connections.

Also adds `Item::from_name` (typed inverse of `Item::name`).

## Checklist

`cargo fmt`, `cargo clippy` (default + `--no-default-features --tests`) `-D warnings`, and `cargo test --no-default-features` (89 tests) all pass. Python side is untouched (test-only Rust + YAML data).


beyarkay:
## Addressed code-review (simplifications, linter avoidances, robustness)

A multi-agent review focused on simplifications, unnecessary lint suppressions, and non-robust shortcuts. No correctness bugs were found in the core parser; the changes below are the actionable items.

**Linter avoidance**
- `Item::from_name` → `#[cfg(test)]` instead of `#[allow(dead_code)]`. It's only used by the test-only `textual` module, so it no longer ships as release API and needs no suppression. (The two `#[allow(clippy::needless_range_loop)]` and the tests-module `unwrap`/`expect` allows were reviewed and kept — genuinely required.)

**Robustness**
- `parse_many` numbers errors by **document position** (`enumerate`), so a skipped empty document no longer shifts the reported factory index in failure messages.
- `place_multi` now requires every footprint tile's direction to agree (`None` for square entities), rejecting a stray marker like `a>` instead of silently swallowing it.
- New guard test `test_entity_registry_matches_placeable_items`: adding a placeable `Item` to the enum without an `ENTITY_CHARS` entry now **fails loudly** (previously it silently rendered as `?` and was unparseable).

**Simplifications**
- Extracted shared `bbox()` / `on_perimeter()` helpers (were duplicated verbatim in `place_multi` and `render_multi`).
- Dropped the redundant `seen_dir` flag in `classify` (`dir != None` is equivalent).
- Removed `char_for_item` — it's `render_char(item, Misc::None)`.
- Trimmed `from_name` spot-check tests already covered by the roundtrip loop (per the repo's "don't re-test static data" convention).

90 Rust tests pass; `cargo fmt`, `clippy` (default + `--no-default-features --tests`) `-D warnings`, and the Python suite are green.

PR #222, opened 2026-07-03
Mask illegal tiles in greedy rollout eval

beyarkay:
The greedy rollout eval argmaxes the tile head, which livelocks the moment
its top choice is an occupied or unbuildable tile: it re-proposes the same
illegal tile every step, the env rejects it, and the rollout burns to
max_steps with no throughput. Restrict the tile argmax to legal placement
targets — tiles that are empty (ENTITIES == empty) and buildable
(FOOTPRINT == AVAILABLE), the same pre-checks FactorioEnv.step runs before
rejecting a placement.

The mask is computable straight from the observation, so a deployed policy
can apply the identical mask; the masked throughput is thus the honest
estimate of deployable skill. Masking is gated on a new mask_illegal_tiles
flag (default on) and applied only in eval — training is untouched. PPO's
_run_greedy_eval reuses run_rollout_eval, so it picks up masking for free.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01ErodEfsjviZg3z3zo4FMdR

### DIFF
--- a/ppo.py
+++ b/ppo.py
@@ -72,6 +72,9 @@ _CH_DIR = Channel.DIRECTION.value
 _CH_ITEMS = Channel.ITEMS.value
 _CH_MISC = Channel.MISC.value
 _CH_FOOTPRINT = Channel.FOOTPRINT.value
+# Footprint is two-valued (UNAVAILABLE / AVAILABLE) and exhaustive, so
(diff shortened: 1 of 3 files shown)


beyarkay:
waiting until an SFT training run can be done on this branch before merging

PR #245, opened 2026-07-05
Encode nominal observation channels categorically before the conv

beyarkay:
The observation's ENTITIES/ITEMS/DIRECTION/MISC channels held raw
category ids as single float channels, so the conv saw them as ordinal
magnitudes (belt=1, assembler=3, splitter=5) — meaningless for nominal
labels. Encode each categorically before the encoder instead: learned
embeddings for the two wide vocabularies (entity/item, ~68 each) and
one-hots for the two narrow ones (direction=5, misc=3). FOOTPRINT is
genuinely binary, so it stays a scalar. The output heads were already
categorical (softmax over classes) and are unchanged.

AgentCNN.input_channels replaces len(Channel) as the first conv's input
width, and _encode_input() does the expansion for every forward path
(get_value/eot_prob/get_action_and_value plus the manual forwards in
sft.py's rollout/train loops and factory_builder). This changes the
first conv's shape, so pre-existing checkpoints no longer load.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01YSWn2kjVHx995L8gE2EMT1


beyarkay:
Damn based on [uju9n8ql](https://wandb.ai/beyarkay/factorion/runs/uju9n8ql/) (this PR) vs [u6v4od88](https://wandb.ai/beyarkay/factorion/runs/u6v4od88) (main), this PR is significantly better than HEAD, way way better

PR #272, opened 2026-07-08
Add held-out recipe validation and assembler item accuracy metrics

beyarkay:
## Summary

This PR adds support for held-out recipe validation during SFT training and introduces assembler-specific item-head accuracy metrics to track the model's ability to copy recipe tags onto assembling machines. This enables measuring generalization to unseen recipes vs. memorization of seen ones.

## Key Changes

- **Held-out recipe filtering**: Added `_heldout_recipe_ids()` to deterministically select a fraction of recipes per ingredient-count class to exclude from training while keeping them in validation. Ensures at least one recipe per class remains trainable.

- **Training-time filtering**: Modified `_iter_demo_pairs()` and `_materialise()` to accept `heldout_recipe_ids` parameter and skip factories whose assembler recipe is in the held-out set. Updated `StreamingDemoDataset` to propagate this parameter through the data pipeline.

- **Assembler placement detection**: Added `_factory_recipe_id()` helper to extract the recipe ID from a factory's assembler machine, enabling filtering and metric computation specific to assembler placements.

- **SFT validation metrics**: 
  - Added `val/asm_item_acc` — item-head accuracy restricted to assembler placements only (the recipe-pick signal)
  - Added `val/asm_item_acc_heldout` and `val/asm_item_acc_indist` — splits assembler accuracy into held-out (generalization) vs. in-distribution (memorization) recipes when `--heldout-recipe-frac > 0`
  - Per-kind breakdown: `val/{LESSON_KIND}/asm_item_acc` only emitted for lessons that actually place assemblers (MEMORISE lessons)

- **PPO evaluation**: Added `run_asm_item_acc_eval()` function for teacher-forced recipe-pick accuracy evaluation on held-out factories, mirroring SFT's validation. Integrated into PPO's greedy eval to track `eval/asm_item_acc` and per-kind variants.

- **Configuration**: Added `heldout_recipe_frac` parameter to `SftArgs` (default 0.0, disabled) to control the fraction of recipes held out per ingredient-count class.

## Implementation Details

- Held-out recipes are selected deterministically by sorting recipe names within each ingredient-count class and taking the first `floor(frac * class_size)` recipes, ensuring reproducibility and leaving at least one recipe per class trainable.
- Assembler placements are identified by checking if an entity is `_ASM_MACHINE_ENT_ID` at a placement step (`eot == 0`).
- The item-head accuracy for assembler placements uses the same denominator (assembler count) across all metrics, preventing lessons with no assemblers from diluting the overall accuracy toward 1.0.
- Validation set remains unfiltered to contain both seen and unseen recipes, enabling the heldout/indist split to measure generalization.
- Tests added to verify deterministic held-out set generation, training exclusion of held-out recipes, and correct metric reporting in both SFT and PPO contexts.

https://claude.ai/code/session_01MNwQ7nbwU9ZqksfjNx19xH


beyarkay:
this pr just adds some metrics, let's see what they look like

Issue #261, opened 2026-07-08
Automate parity with factorio


beyarkay:
Right now the Factorion engine is flying blind and via tests in order to get parity with he Factorio engine, but ideally we'd setup a Factorio mod so that Claude can write code adjusting the engine and then verify that the outcome matches Factorio v1. Idk how easy it'll be to do this or even how to do the verification, we want to test that throughput is calculated correctly in a wide variety of cases but we can only measure throughput indirectly in Factorio by building a factory and letting it run.

Maybe best would be a mod that gives Claude control over a local Factorio instance, place down a factory, and then a way to speedup the tick rate so that enough time can pass for all the items to flow through and then finally we can maybe use a mod to easily measure the thput at some final state.

This would only be a binary classifier though, it would say if there's an issue but not *where* the issue is. Idk how best to look for the issue, visually it's pretty clear but Claude isn't good with visuals. Maybe we can calculate the thput/items/blockages on *every* entity in the factory and then diagnose where things halt from there? I think there's a mod that highlights when you entities are waiting on input items or waiting on space in the output so maybe this'll work

PR #277, opened 2026-07-09
Bug: lesson generator emits invalid sink-loop-to-source factories

beyarkay:
**Found by the Factorio parity harness (#261).** Bug report as a failing test — the added test asserts the desired invariant, so it fails until the generator is fixed.

## The bug
`build_factory` can emit a degenerate factory whose **sink is placed facing an occupied tile**, feeding back into the factory. The clearest case is `MOVE_ONE_ITEM` seed 99: the sink sits directly above the source, facing north *into* it, forming a `source → … → sink → source` loop.

```
S> bv
K^ bv     <- sink (2,4) faces north into the source (2,3)
b^ b<
```

The engine models a sink as an **infinite consumer**, so it scores this as a normal 15/s belt and the factory passes generation. But the layout is invalid — in real Factorio a belt on the sink tile outputs its items back into the factory, and the parity harness measures an impossible **135/s** there.

## Scope (origin/main, 300 seeds × all lessons)
- `MOVE_ONE_ITEM`: seeds 99, 241
- `SPLITTER_SPLIT`: seeds 7, 95, 190, 237
- `MOVE_ONE_ITEM_CHAOS`: seeds 99, 241

## The fix
The generator should reject (or repair) any factory where a sink's output tile is occupied — a sink must be a pure consumer. The added test `tests/test_sink_is_pure_consumer.py` asserts exactly this and currently FAILS.

### DIFF
--- a/tests/test_sink_is_pure_consumer.py
+++ b/tests/test_sink_is_pure_consumer.py
@@ -1,0 +1,92 @@
+"""A generated sink must be a pure consumer — it must not feed back into
+the factory.
+
+FAILING TEST documenting a lesson-generator bug found by the Factorio
+parity harness (issue #261). `build_factory` can emit a degenerate factory
+whose sink is placed facing an occupied tile — e.g. MOVE_ONE_ITEM seed 99
+puts the sink directly above the source, facing north *into* it, forming a
+source→…→sink→source loop.
+
+The Factorion engine models a sink as an infinite consumer, so it scores
+such a factory as a normal 15/s belt. But the layout is invalid: in real
+Factorio a belt on the sink tile would output its items back into the
+factory (the parity harness measures an impossible 135/s there). The
+generator should never produce it.
+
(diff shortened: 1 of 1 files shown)


beyarkay:
Converted to #429 and closing this.

The bug is **still live** — I re-ran this test against `main@1feaf7a` and it fails. The scope has grown since July: 46 violations across 7 lesson kinds (was 3 kinds), with the MEMORISE lessons now the worst affected and the rate climbing with ingredient count (0.7% at 1 ingredient → 6.7% at 4). The original `MOVE_ONE_ITEM` seed 99 factory still reproduces exactly as described above.

Closing only because a test-only PR can never merge green — the finding is preserved in #429 with the current scope table, three worked examples (sink→source, sink→inserter, sink→sink), and this test named as the acceptance criterion. The branch stays reachable, so the test can be cherry-picked straight into whichever PR carries the generator fix.

---

PR #290, opened 2026-07-13
Add global-context feature to per-tile heads

beyarkay:
Give the per-tile ent/dir/item/misc heads grid-global context. They read a single feature column `encoded[:, :, x, y]`, so their receptive field is capped at the conv window — but recipe choice, underground-belt spans, and "which way is the sink" are grid-global questions. Pool the whole encoded map (mean+max over the spatial axes, projected to `global_feat_dim=32`) and concatenate it onto the tile features before the four heads. `global_feat_dim=0` recovers the window-only heads.

Wired at every direct head callsite so a checkpoint from this arch loads everywhere:
- `ppo.py` `AgentCNN` (`get_action_and_value`, the `_global_feat` helper, head input-dim bump)
- `sft.py` train / val / greedy-rollout forward paths
- `scripts/factory_builder.py` prediction UI and `factorion-mod/server/server.py` mod inference server
- `tests/test_sft.py` convergence test updated for the new head input dim

## Result (controlled MEMORISE_2 compare)

Measured apples-to-apples on MEMORISE_2-only data (same lesson, same val split, differing only by this arch), seed-paired, this branch vs the windowed baseline:

| Metric | windowed | global-ctx | Δ | p |
|---|---|---|---|---|
| `val/MEMORISE_2_INGREDIENT_RECIPES/asm_item_acc` | 0.072 ± 0.007 | **0.869 ± 0.031** | +0.797 | 0.014 |
| `val/thput` | 0.053 | **0.949** | +0.897 | 0.011 |
| `val/thput_eot` | 0.054 | **0.985** | +0.931 | 8.5e-4 |
| `val/item_acc` (recipe head) | 0.866 | **1.000** | +0.134 | 0.002 |

Recipe selection (`asm_item_acc`) goes from ~0.07 → ~0.87 and built-factory throughput ~0.05 → ~0.95 — the receptive-field-starvation fix H1 predicted. Effect sizes are large; the compare was effectively n=2/side (one baseline seed's pod never produced a run), so the exact p-values are soft but the deltas are far beyond that noise.

Generalization to the other assembler lessons (the full 11-lesson curriculum) is a separate follow-up; this PR is the arch change alone.

### DIFF
--- a/factorion-mod/server/server.py
+++ b/factorion-mod/server/server.py
@@ -142,6 +142,8 @@ def _argmax_action(agent: AgentCNN, obs_CWH: np.ndarray, device) -> dict:
         x_B = tile_idx // agent.height
         y_B = tile_idx % agent.height
         feat = encoded[torch.arange(B), :, x_B, y_B]
+        if agent.global_feat_dim > 0:
+            feat = torch.cat([feat, agent._global_feat(encoded)], dim=1)
         ent = agent.ent_head(feat).argmax(dim=-1)
         dir_ = agent.dir_head(feat).argmax(dim=-1)
         item = agent.item_head(feat).argmax(dim=-1)
--- a/ppo.py
+++ b/ppo.py
@@ -1131,7 +1131,7 @@ def _categorical_entropy(logp_all_BN):
 
 
 class AgentCNN(nn.Module):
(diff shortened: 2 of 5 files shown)


beyarkay:
/ci compare sft --num-samples 20000000 --seeds 3 --base-ref claude/train-memorise2-only

beyarkay:
/ci compare sft --num-samples 20000000 --seeds 3

Issue #161, opened 2026-06-19
Throughput normalization depends on the scripted full factory (breaks for arbitrary RL rollouts)


beyarkay:
## Context

We normalize an episode's throughput by the **per-factory maximum** so that a perfectly-built factory scores `1.0` and an empty one scores `0.0`, replacing the old hardcoded `/ 15.0` (which mislabeled any factory whose max throughput was < 15 — a perfect build could score e.g. 0.33).

The max is computed in `FactorioEnv.reset()` as `simulate_throughput(self._solved_world_CWH)`, i.e. it relies on having the **complete, correct reference factory** that `build_factory()` produced for the scripted lesson.

## The problem (fine for now, must change later)

This dependence on the full scripted solution is acceptable today because **every** RL/SFT episode is a scripted lesson, so we always have the reference factory and therefore its max throughput.

It will **not** hold once we do arbitrary RL rollouts that aren't scripted lessons (no known-optimal reference factory to normalize against). At that point:

- `thput_normed` (`raw / max`) can't be computed — there is no `max`.
- The episode **termination** condition (`thput_normed >= 1.0`, i.e. "reached the reference max") is undefined.
- Any reward/curriculum logic that assumes throughput ∈ [0, 1] needs revisiting.

## What's in place now

- `info['thput_raw']` — raw items/second (reference-free; this is what arbitrary RL should ultimately optimize).
- `info['thput_normed']` — `raw / max` clamped to [0, 1] (depends on the scripted reference).
- SFT's `val/throughput` uses `thput_normed`; PPO reward/termination/curriculum use `thput_normed`; PPO also logs `thput_raw`.

Each max-dependent site is marked with a `FIXME` pointing at this issue.

## When we tackle arbitrary RL rollouts

Decide on a reference-free signal, e.g.:
- optimize `thput_raw` directly (items/sec), and
- replace the "reached max" termination with a step budget / explicit stop action (the SFT EOT head already does this).

PR #297, opened 2026-07-15
Add TRIAL_RECIPE_TREE_DEPTH_1..3: RL-only trials with no known solution

beyarkay:
## What

**1. Trial-kind infrastructure** — some kinds may now have *no known solution*. A **trial** puts the agent in a scenario (markers only) and it must figure the rest out through RL reward; there is nothing to imitate.

- `LessonKind::is_trial()` in Rust, exposed via `py_lesson_is_trial()` and mirrored as `LESSON_IS_TRIAL` in `factorion.py`.
- Every `BuiltFactory`/`Factory` now carries `max_throughput` — the items/s reference throughput is normalized by: the solved world's simulated rate for scripted lessons, an **analytic ceiling** for trials. `FactorioEnv.reset` reads it off the factory instead of re-simulating the solved world (first bite out of #161: trials have no reference solution to simulate).
- The SFT demo sampler skips trial kinds — they can never emit an expert pair, and the fewest-pairs draw would otherwise redraw them forever. PPO's random-kind sampling and `eval/` set include them as usual, so per-trial curves show up alongside the lessons.

**2. `TRIAL_RECIPE_TREE_DEPTH_1/2/3`** — pick a random craftable item as the **sink**, walk its ingredient tree randomly expanding expandable ingredients, and place one **source** per unique frontier item. Nothing else is placed. E.g. an `electronic-circuit` sink gets sources {copper-cable, iron-plate} (no expansion) or {copper-plate, iron-plate} (cable expanded).

- Depth-split like the MEMORISE family: expansion is capped at the kind's depth and the draw is rejected unless the deepest leaf lands *exactly* on it — `DEPTH_1` is exactly the direct ingredients, `DEPTH_2/3` are progressively harder rungs with their own per-kind metrics.
- `max_throughput` = the sink recipe's single fully-fed assembler output rate (the engine's one-machine ceiling under `transform_flow`), so `thput_normed` stays a meaningful [0, 1] signal without a reference factory.
- `total_entities == 0`: `blank_entities` is a no-op and every episode starts from the markers.

**3. Any assembler tier** — the throughput engine's `AssemblingMachine` crafts whatever recipe it's tagged with and never checks `produced_by`, so trial sinks/expansion aren't restricted to AM1-craftable recipes (`engine-unit` is eligible).

**4. Edge placement** — markers always sit on the grid edge working inward: sources face their wall's inward normal (dropping onto the interior); the sink faces outward, since a sink pulls from the cell behind it — so every marker's working cell is an interior neighbour and each is reserved exclusively for that marker.

**5. Cleanup passes** — stale pre-PR comments dropped (`/15.0` normalization, the `FIXME(#161)` that this PR resolves for generated factories); trial builder simplified; test scaffolding deduplicated (one shared tensor scan on the Python side, one shared edge-marker invariant check on the Rust side).

## Tests

- Rust: `test_recipe_tree_trial_smoke` (markers-only worlds, frontier validity via a cover check, exact-depth chain reachability, depth-1 == direct ingredients, analytic ceiling, edge/orientation + working-cell invariants); the no-orphan test skips trials (their markers are unconnected by construction).
- Python: `tests/test_recipe_tree_trial.py` (trial flags, blank no-op, env episode starts from markers with positive normalization ceiling, SFT stream never draws a trial, frontier validity from the `recipes` dict, edge/orientation invariants).
- Two existing tests updated: "every kind contributes pairs" → every *non-trial* kind; one env-reset test now pins a scripted kind since a randomly-sampled kind can be a trial with nothing to blank.

Pre-completion checklist: cargo fmt/clippy/test (155 passed), maturin build, full pytest (3099+ passed), ruff, ty (same 5 pre-existing diagnostics as main), PPO + SFT smoke tests all green.

https://claude.ai/code/session_01Ddn88wY8wpVMDGVnkq6Pet


beyarkay:
Superseded by #344, which rebases this branch onto current `main` (this one was 20 commits behind and predated the `thput_eot`→`thput` rename, `SPLITTER_MERGE_SIDELOADED`, EOT-skips-placement, action masking and the attention arch) and adds the lesson/trial metric split — `eval/thput` and the `rollout/*` aggregates stay lesson-only, trials report under `eval/trial_thput` and `rollout/trial_*`.

All the work here is carried over. Closing as a duplicate.

---

PR #313, opened 2026-07-17
Sweepable global-knowledge architecture knobs + arch sweep config

beyarkay:
Follow-up to #290 (mean+max global vector, validated only on MEMORISE_2): make "how does the model get grid-global knowledge" a sweepable axis instead of a single hardcoded mechanism. All mechanisms are encoded as **zero-off numeric knobs** on `SharedArgs` (same trick as the `layer1..8` width slots) because W&B Bayes has no conditional parameter spaces and the mechanisms compose — they are not mutually exclusive architectures. Int flags rather than bools throughout, because W&B sweeps pass `--param=value`, which tyro bools reject.

## The knobs

| Flag | Default | Mechanism |
|---|---|---|
| `--global_feat_dim` | 32 | #290's pooled mean+max global vector, now a CLI flag (was hardcoded); 0 disables |
| `--global_broadcast` | 0 | broadcast the global vector as constant channels onto every cell of the encoded map, so the **tile head** (which picks *where* to act, first in the action factorization, previously window-only) and the flatten critic/eot heads see global context. Per-tile head inputs are provably identical between modes (tested) |
| `--coord_channels` | 0 | CoordConv (Liu et al. 2018): two normalized x/y input channels — a translation-equivariant conv stack otherwise can't express absolute position |
| `--dilation_growth` | 1 | conv layer *i* gets dilation *d*^*i* (Yu &amp; Koltun 2016): exponential receptive-field growth at zero parameter cost |
| `--se_ratio` | 0 | squeeze-and-excitation gating (Hu et al. 2018) after each conv layer: per-layer global channel modulation |
| `--global_critic` | 0 | critic + eot heads read the pooled global vector (`Linear(G,1)`) instead of `Flatten→Linear(C·W·H,1)` — far fewer params, and the global pathway gets gradient from the value/eot losses |

Encoder depth (`layer4..8`) and `kernel_size` were already sweepable; the sweep config now actually uses them.

## Refactor

All `AgentCNN` consumers (PPO forward, SFT train/val/rollout, factory-builder UI, mod inference server) now go through shared trunk helpers `encode()` / `tile_features()` / `critic_value()` / `eot_logit()`, replacing six hand-composed copies of the `encoder → global-concat → heads` sequence — an arch variant now changes one place. Incidentally fixes `server.py:_argmax_action` feeding the raw observation to the encoder without `_encode_input`.

## Sweep

`ci/sweep_sft.yaml` becomes the arch sweep: the arch knobs + `lr` (its optimum moves with the arch); the previously-swept training hpars fall through to their tuned `training_config.py` defaults. Metric switches `val/acc` → **`val/thput`** — #290 showed per-head accuracy sitting high while built-factory throughput was ~0.05, and `val/thput` is already the checkpoint-selection metric. Sweeping at the SFT stage (not PPO) since arch changes require training from scratch.

Genuinely incompatible arch families (self-attention over the 121-cell map) are deliberately **not** in this PR — one PR per incompatible family, each editing `sweep_sft.yaml` in place, so the CI never needs to resolve which YAML to sweep.

## Not done / known limitations

- `scripts/factory_builder.py` infers only depth/width/kernel from checkpoint conv shapes; checkpoints trained with `global_broadcast`/`se_ratio`/`coord_channels`/`dilation_growth` on won't reconstruct there (clear shape-mismatch error, not silent). Worth extending once the sweep picks a winner.
- Merging the four per-tile heads into one spliced `Linear` was considered and skipped: output rows of a Linear are independent, so it is mathematically identical to the current four heads — pure churn with nothing to sweep.

## Testing

- New `TestArchVariants` suite: every knob (and the everything-on combo) through full PPO forward + stored-action recompute + state-dict roundtrip; broadcast≡concat equivalence for the per-tile heads; CoordConv plane contents; dilation schedule + grid-size invariant; SE shape/arch-inference safety; global-critic param counts and `head[-1]` reachability.
- Full checklist: cargo fmt/clippy/test, 3122 Python tests, ruff, ty, PPO + SFT smoke tests — all green.
- Extra: SFT and PPO smoke runs with **all knobs on** (via the exact underscore `--param value` form the sweep agent uses), plus an SFT→PPO `--start-from` handoff of an all-knobs-on checkpoint.

https://claude.ai/code/session_01McAdC7meCg1kdbgCHKkKDu

---
_Generated by [Claude Code](https://claude.ai/code/session_01McAdC7meCg1kdbgCHKkKDu)_


beyarkay:
Superseded by #314, which is now standalone (rebased onto `main`) and carries forward the parts of this PR worth keeping — the `encode()`/`tile_features()`/`critic_value()`/`eot_logit()` shared-trunk helpers, CoordConv, and the CI fixes (`agents_per_pod=1` on both launch paths, 24h sweep-pod deadline, widened CUDA versions).

The zero-off knob matrix this PR added did its job: the follow-up sweep (`sn7fd8l8`, 60 runs) showed the self-attention stage + CoordConv + the pooled global vector are the winning combination (`val/thput_eot` 0.85 vs ~0.51), while `se_ratio`, `dilation_growth`, `global_broadcast` and `global_critic` were all dead. #314 keeps only what won and removes the rest, so maintaining this larger knob surface in parallel isn't worth it. Closing in favour of #314.

---

PR #317, opened 2026-07-20
Run CI on CUDA 12.6 torch to widen the schedulable GPU pool

beyarkay:
## Why

CI could only schedule onto hosts advertising CUDA 13.0–13.3, which is a small
slice of the RunPod fleet — a frequent source of failed launches. That
requirement was never a real one: it came from `uv.lock` resolving `torch` from
default PyPI, whose 2.12.x wheel is the `+cu130` build, bundling a CUDA 13
runtime that needs host driver >= 580. `ci/config.py` then narrowed
`ALLOWED_CUDA_VERSIONS` to match. Nothing in the code needs CUDA 13.

## What

`torch==2.12.1+cu126` exists on the PyTorch index, so the **torch version is
unchanged** — only its CUDA build moves. cu126 runs on any 12.x driver via CUDA
minor-version compatibility, and 13.x hosts stay eligible because drivers are
forward-compatible with older runtimes. `ALLOWED_CUDA_VERSIONS` goes from 4
entries to 14.

Every GPU in the fallback chain is Ampere/Ada (sm_86/sm_89), all covered by CUDA
12.6. A Blackwell card would need cu129+ — noted in `pyproject.toml`.

Supporting changes:

- **torch pinned to `2.12.1`.** The cu126 index carries 2.13.0 while PyPI does
  not, so an unpinned resolve put Linux on a version macOS cannot install.
- **`Dockerfile.ci-gpu` passes the PyTorch index explicitly.** `uv export` drops
  index URLs, so `2.12.1+cu126` would otherwise resolve against PyPI and fail.
  Raised `UV_HTTP_TIMEOUT` and added a cache mount because the `nvidia-*` wheels
  come from `pypi.nvidia.com`, which stalls past uv's 30s default (this failed
  the first build).
- **Image tagged `:cu126`, not `:latest`,** so a commit's `DOCKER_IMAGE` names an
  image that cannot drift after it was tested.
- **CUDA build, host driver, and compute capability logged to W&B config**, plus
  the driver version in the pod log. CI schedules onto whatever driver is free,
  so a host-specific failure was previously untraceable once the pod was gone.

## Testing

Locally: `ruff`, `ty`, and the full pytest suite (3096 passed, 708 skipped) are
green; the SFT smoke test passes. The **PPO smoke test fails on macOS** — torch
inductor cannot compile its Metal shaders under MPS. This is pre-existing and
unrelated: `main`'s `ppo.py` fails identically on the same machine.

The real test is a `/ci sft` run on this branch, which is what actually
exercises the cu126 image on a RunPod host. Verifying via the `env/cuda_driver`
value the run reports.

### DIFF
--- a/Dockerfile.ci-gpu
+++ b/Dockerfile.ci-gpu
@@ -1,9 +1,11 @@
-# GPU image for RunPod CI/training. Build + push by hand when deps change (no CI job):
-#   docker build --platform linux/amd64 -f Dockerfile.ci-gpu -t beyarkay/factorion-ci-gpu:latest . \
-#     && docker push beyarkay/factorion-ci-gpu:latest
+# GPU image for RunPod CI/training. Build + push by hand when deps change (no CI job).
+# Tag by CUDA build rather than :latest, so ci/config.py's DOCKER_IMAGE names an
+# image that can't drift underneath a commit that was tested against it:
+#   docker build --platform linux/amd64 -f Dockerfile.ci-gpu -t beyarkay/factorion-ci-gpu:cu126 . \
+#     && docker push beyarkay/factorion-ci-gpu:cu126
 #
 # Slim base (python:3.11-slim, ~150 MB) instead of runpod/pytorch's CUDA-devel
-# image: torch 2.12.x+cu130 (pinned in uv.lock) bundles its own CUDA 13 runtime,
+# image: torch 2.12.x+cu126 (pinned in uv.lock) bundles its own CUDA 12.6 runtime,
 # so the base needs no CUDA toolchain — the RunPod host injects the GPU driver at
 # run time. runpod_start.sh replicates RunPod's SSH bring-up so direct-SSH CI
 # keeps working. --platform pins amd64 so a build from an arm64 Mac still targets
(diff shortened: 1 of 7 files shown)


beyarkay:
### Confirmed on a CUDA 12.4 host ✅

SFT run [`fxy6gh4e`](https://wandb.ai/beyarkay/factorion/runs/fxy6gh4e) at `9efa189` finished on pod `ad9gej1bhnjq2g`.

| | |
|---|---|
| Host driver CUDA (`cudaVersion`, W&B metadata) | **12.4** |
| torch | `2.12.1+cu126` |
| GPU | NVIDIA RTX A4000, compute capability 8.6 (Ampere) |
| Samples seen | 50,000 |
| `val/acc` / `val/ent_acc` / `val/eot_acc` | 0.049 / 0.834 / 0.904 |

**12.4 is the result that matters:** `main`'s `ALLOWED_CUDA_VERSIONS = ["13.0", "13.1", "13.2", "13.3"]` would have refused to schedule this pod at all. The cu126 build ran on it via CUDA minor-version compatibility, which is the mechanism this PR depends on — so the reduced requirement is exercised end to end, not just theoretically sound.

The pod also came up on the first attempt, with no `This machine does not have the resources to deploy your pod` retries. That is a single sample, so it is suggestive rather than proof, but it is consistent with a larger eligible host pool (4 → 14 accepted CUDA versions).

One bug this run caught: `env/cuda_driver` was silently missing from the logged config, because `torch._C._cuda_getDriverVersion` does not exist in torch 2.12 and the `getattr` fallback made the branch unreachable. Removed in `b23fc44` — wandb already records the host driver as `cudaVersion` in run metadata, which is where the 12.4 above came from.

Issue #323, opened 2026-07-22
New lesson: SPLITTER_MERGE_DIFF_ITEMS — a throughput-honest merge via two distinct items


beyarkay:
## Motivation

`SPLITTER_MERGE` is reward-hackable. Both `SPLITTER_SPLIT` and `SPLITTER_MERGE` come from one generator (`build_splitter_y` in `factorion_rs/src/factory_gen.rs`): a base wired through a splitter to two prongs, with `SPLITTER_MERGE` just reversing the flow (2 sources → splitter → 1 sink).

The problem is scoring. `factory_score` (`factorion_rs/src/throughput.rs`) is a power-mean over per-**sink** deliveries, and per-factory throughput is normalized as `raw ÷ reference-factory-max` (`ppo.py`). MERGE has a **single** sink, and that sink can only ingest **one belt = 15/s**, which **one** source already saturates. So the second source's entire arm is throughput-redundant.

Measured over 100 generated factories at size 11 (the training size), dropping one arm:

| Lesson | `thput_normed` with one arm missing | reaches 1.0 |
|---|---|---|
| SPLITTER_SPLIT | mean 0.22, median 0.25 | 0% |
| SPLITTER_MERGE | mean 0.86, median 1.00 | 86% |

This is why `val/SPLITTER_MERGE/thput_eot` plateaus at ~1.0 almost immediately while `SPLITTER_SPLIT` sits at ~0.7 — MERGE's metric rewards building only half the factory (and arguably not even the splitter), so it never actually teaches merging.

A companion lesson `SPLITTER_MERGE_SIDELOADED` (being built separately) fixes this by side-loading each source arm down to 7.5/s so both are needed to fill the 15/s output. This issue tracks the **other** honest-merge variant: make each source carry a **different item** that the sink demands, so both arms are throughput-necessary by construction.

## Goal

`SPLITTER_MERGE_DIFF_ITEMS`: two sources emit two **distinct** items (e.g. iron-plate and copper-plate). They merge through a splitter onto a single output belt, which carries both items (one per lane). The demand side requires **both** items, so omitting either source arm is penalized — the merge becomes genuinely load-bearing without any side-load trick, and with **no orphan tiles** (unlike the sideloaded variant).

Because each item rides its own belt lane, one output belt legitimately carries `7.5 (itemA) + 7.5 (itemB) = 15/s`. A build that omits one source arm delivers only one item → the missing item's demand scores 0 → the power-mean drops it well below 1.0.

## Design

[...]

PR #325, opened 2026-07-22
Replace SPLITTER_MERGE with throughput-honest SPLITTER_MERGE_SIDELOADED

beyarkay:
## Why

Investigating why `val/SPLITTER_SPLIT/thput_eot` plateaus at ~0.7 while `SPLITTER_MERGE` sits at ~1.0 (an almost-mirror lesson) turned up a reward hack in **MERGE**, not a problem in SPLIT.

Both lessons came from one generator (a splitter-Y with the flow reversed for merge). But scoring is a power-mean over per-**sink** deliveries, and MERGE has a **single** sink. That sink can only ingest one belt (15 i/s), which **one** source already saturates — so the second source's entire arm is throughput-redundant. Measured over 100 size-11 factories, dropping one arm:

| Lesson | `thput_normed` with one arm missing | reaches 1.0 |
|---|---|---|
| SPLITTER_SPLIT | mean 0.22, median 0.25 | 0% |
| SPLITTER_MERGE (old) | mean 0.86, median 1.00 | 86% |

MERGE's ~1.0 was the model learning to build **half** the factory (and not necessarily even the splitter). SPLIT's ~0.7 is honest difficulty. The env isn't broken — the *metric on MERGE was degenerate*.

## What

Replace `SPLITTER_MERGE` with **`SPLITTER_MERGE_SIDELOADED`**, a merge that is throughput-honest by construction:

- Each source is capped to **7.5 i/s** by a **side-load gadget** — a protected empty *decoy* belt is the permanent second belt-input that forces the source to side-load onto one lane (7.5) instead of curving to a full 15. (A lone side feed reverts to 15 — verified — which is exactly why the decoy must be protected and always present.)
- A splitter merges the two 7.5 arms into one 15 i/s belt to a single sink, so **both arms are throughput-necessary**: drop either and the sink falls to 7.5.
- Both wiki side-load shapes are generated (opposing `S> bv <decoy` and inline decoy-above).

## No-orphan invariant

The decoy belt carries nothing and lies on no source→sink path, so it is an intentional orphan. The no-orphan invariant gains a single per-lesson opt-out — `LessonKind::allows_orphans()`, exposed to Python as `factorion_rs.py_lesson_allows_orphans()`. The test then requires every orphan to be a **protected** tile (`unreachable == protected_positions.len()`), so no *accidental* orphan slips through.

## Notes / limitations

- Requires grid `size ≥ 7` (the old merge worked at ≥ 5). The samplers already skip `None`-producing (size, kind) combos, so on tiny grids (e.g. the size-5 SFT smoke) this lesson is simply excluded from the shard — confirmed, no crash.
- The now-dead reverse-flow machinery in the splitter-Y generator (`place_plan`/`place_run`/`build_splitter_y`'s `merge` path) is removed; `build_splitter_split` is the sole caller.
- Companion lesson **`SPLITTER_MERGE_DIFF_ITEMS`** (the orphan-free alternative — two sources emit different items a single sink demands) is filed as #323 for a follow-up; together the two replace the original hackable merge.

## Tests / checklist

New coverage: `factory_gen.rs::tests::test_sideloaded_merge_requires_both_arms` (full build = 15, exactly 2 protected decoy orphans, blanking either source collapses the sink) and Python `TestSplitterMergeSideloadedHonest` (same anti-hack property at the training-facing layer, plus `py_lesson_allows_orphans`). All existing splitter tests were migrated to the new name.

Full pre-completion checklist green: Rust fmt + clippy (`-D warnings`) + 158 tests, `maturin develop`, 3116 pytest, ruff, ty, PPO smoke (5k steps), SFT smoke.

https://claude.ai/code/session_01XJneofoFE64TrPun34kgTQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01XJneofoFE64TrPun34kgTQ)_


beyarkay:
/ci compare sft --num-samples 1000000 --seeds 1

PR #327, opened 2026-07-22
Mask invalid action combinations

beyarkay:
## Summary

- condition direction, item, and misc logits on the selected entity in the shared sampler
- force rotation-invariant assemblers to use `direction=NONE` and restrict their item choice to recipes supported by assembling machine 1
- force empty placements, belts, inserters, splitters, and underground belts to use only their valid field combinations
- mask non-placeable item IDs from the entity distribution without changing the entity-head shape
- apply the same masks when PPO recomputes stored-action log-probabilities
- keep semantic masks out of checkpoints so existing checkpoints remain loadable

## Why

The policy heads were sampled independently, so an otherwise correct placement could be rejected because an irrelevant head emitted an incompatible value—for example, a belt with a recipe item or an underground belt with `misc=NONE`. These small per-head errors became invalid whole actions and compounded over long factory rollouts.

The new masks encode only the action grammar already enforced by the environment. They do not reveal the target layout or recipe; the model still chooses the tile, entity, meaningful direction, underground endpoint, and assembler recipe.

Masked logits are also used by PPO replay so sampled actions and recomputed log-probabilities remain consistent. Entropy calculation now handles masked `-inf` log-probabilities without producing NaNs.

## Validation

- `uv run pytest -q` — 3,125 passed, 708 skipped
- `uv run pytest tests/test_torch_compile.py tests/test_ppo_logging.py -q` — 29 passed
- `uv run ty check ppo.py tests/test_spatial_agent.py`
- `uv run ruff check ppo.py tests/test_spatial_agent.py`

This PR does not change environment rewards or EOT semantics.


beyarkay:
/ci compare sft --num-samples 2500000 --seeds 1

beyarkay:
/ ci compare ppo --start-from x1xzfxia --total-timesteps 2000000 --seeds 1

beyarkay:
ah damn `main` vs `pr` wasn't fair because `main` had #325 which made it sligthly harder to learn things proplery, re-running CI

/ci compare sft --num-samples 2500000 --seeds 3

beyarkay:
/ci compare sft --num-samples 2500000 --seeds 3

beyarkay:
SFT looks ~the same, trying PPO next

beyarkay:
/ci compare ppo --total-timesteps 50000000 --seeds 3 --start-from h3o07we2

beyarkay:
Accidentally did ~50x more steps than needed. There was ~zero difference, maybe a slight improvement, so merging

PR #329, opened 2026-07-23
Weight SFT samples by missing fraction

beyarkay:


beyarkay:
this just.. didn't work. not a budge. didn't have any effect. not good nor bad. ah well.

PR #333, opened 2026-07-24
Use cost-adjusted raw throughput for PPO rewards

beyarkay:
## Summary

- change PPO terminal reward to `thput_raw / (1 + entity_cost_scale * entity_cost)`, with `entity_cost_scale` (η) defaulting to `1e-3`
- calculate each placed entity's cost from its recursively expanded raw-material quantities plus cumulative crafting time, divided by recipe output quantity
- log overall and per-lesson `entity_cost` and `cost_efficiency`
- remove the unused zero step penalty and unit throughput reward scale
- replace the general PPO sweep with a four-run grid that varies only η: `0`, `1e-4`, `1e-3`, and `1e-2`

## Stack

This PR is stacked on #335, which restores the canonical recipes required by entity costing. Its diff contains only the PPO reward, metrics, tests, documentation, and η sweep.

## Why

Raw throughput will not always have a known normalization ceiling. Multiplying it by a bounded efficiency factor keeps throughput primary, makes entity cost secondary, and guarantees that a zero-throughput factory receives zero reward rather than a negative incentive to build nothing.

The grid sweep keeps code, recipes, checkpoint, and PPO hyperparameters identical across runs so `/ci sweep ppo` isolates only η.

## Validation

- `3183 passed, 708 skipped` in the full Python test suite
- `cargo test --manifest-path factorion_rs/Cargo.toml --no-default-features`
- `ruff check .`
- `ty check .`
- exact YAML validation of the four-value η grid


beyarkay:
/ci compare ppo --start-from h76h80yb --total-timesteps 1000000 --seeds 1

beyarkay:
didn't mean to do that..

PR #335, opened 2026-07-24
Restore canonical recipes and balance memorise lessons

beyarkay:
## Summary

- restore all canonical recipes that had been omitted to balance the MEMORISE curriculum
- keep `MEMORISE_N_INGREDIENT_RECIPES` equally difficult by taking the first K eligible recipes from each 1–4 ingredient bucket, where K is the smallest bucket size
- add invariants requiring every recipe key to name a positive output and every agent-placeable entity to have a canonical recipe

## Why

The canonical recipe registry should describe the domain rather than encode curriculum sampling policy. Omitting valid recipes made recursive entity costing incomplete and required special cases outside the registry.

Curriculum balancing now happens explicitly in the MEMORISE sampler while the full recipe registry remains available to every other consumer.

## Validation

- `3183 passed, 708 skipped` in the full Python test suite
- `cargo test --manifest-path factorion_rs/Cargo.toml --no-default-features`
- `ruff check .`
- `ty check .`
- `cargo fmt --manifest-path factorion_rs/Cargo.toml --check`


beyarkay:
/ci compare sft --num-samples 2500000 --seeds 3

beyarkay:
the seeds are identical enough for me

PR #336, opened 2026-07-25
Log-compress the terminal reward across lesson scales

beyarkay:
## Why

Lesson ceilings differ by orders of magnitude in items/s, so an uncompressed terminal reward gives belt lessons ~100x more gradient than assembler lessons, and leaves the small end below a shared critic's error floor. Measured on solved factories (`num_missing_entities=0`, eot immediately, seed 7, size 11) — "share" is each lesson's reward as a fraction of the largest, i.e. roughly its slice of the gradient budget:

| lesson | thput (i/s) | cost | reward (main) | share | reward (PR) | share |
|---|---:|---:|---:|---:|---:|---:|
| MOVE_ONE_ITEM | 15.000 | 20.0 | 14.706 | 100.0% | 7.294 | 100.0% |
| SPLITTER_MERGE_SIDELOADED | 15.000 | 70.8 | 14.009 | 95.3% | 7.246 | 99.3% |
| MOVE_VIA_UG_BELT | 15.000 | 45.0 | 14.354 | 97.6% | 7.270 | 99.7% |
| MOVE_ONE_ITEM_CHAOS | 15.000 | 32.0 | 14.535 | 98.8% | 7.282 | 99.8% |
| CROSS_UNDER_BELT | 15.000 | 67.0 | 14.058 | 95.6% | 7.249 | 99.4% |
| SPLITTER_SPLIT | 7.500 | 66.8 | 7.031 | 47.8% | 6.557 | 89.9% |
| FACTORY_1_INGREDIENT | 4.378 | 242.5 | 3.523 | 24.0% | 5.867 | 80.4% |
| MEMORISE_1_INGREDIENT_RECIPES | 0.860 | 52.8 | 0.817 | 5.6% | 4.415 | 60.5% |
| MEMORISE_4_INGREDIENT_RECIPES | 0.172 | 82.0 | 0.159 | 1.1% | 2.827 | 38.8% |
| MEMORISE_3_INGREDIENT_RECIPES | 0.086 | 72.2 | 0.080 | 0.5% | 2.200 | 30.2% |
| MEMORISE_2_INGREDIENT_RECIPES | 0.043 | 62.5 | 0.040 | 0.3% | 1.619 | 22.2% |
| **spread (max/min)** | | | **363x** | | **4.5x** | |

The gap is a cliff, not a gradient: the five 15 i/s belt lessons already sit within 5% of each other, then SPLITTER_SPLIT halves and the MEMORISE lessons fall to 0.3–5.6%. Note MEMORISE ordering is non-monotonic in ingredient count (1 → 0.860, 2 → 0.043, 3 → 0.086, 4 → 0.172) — that's recipe crafting-time structure, not difficulty, and those four lessons span 20x among themselves before compression.

## What changed

One line in `FactorioEnv.step`'s terminal reward:

```python
reward = thput_raw * cost_efficiency          # unchanged
if self.reward_symlog_r0 > 0:
    reward = math.log1p(reward / self.reward_symlog_r0)
```

with `r0 = PpoArgs.reward_symlog_r0 = 0.01` items/s. Everything downstream — GAE, returns, critic, advantages — then lives consistently in compressed space.

- **Why `log1p` and not plain symlog.** Symlog (`sign(x)·ln(1+|x|)`) is ~identity below 1, so it compresses the top and leaves the bottom untouched: it would take the spread to ~66x, not 4.5x. `r0` moves the knee down to the smallest flow worth caring about; symlog is the `r0 = 1` special case, and rewards here are non-negative so the symmetric branch never applies.
- **`r0` is a units choice, not task knowledge.** One global constant meaning "one item per 100 seconds is the smallest flow we care about". Task-independent, never consults a lesson ceiling or a recipe value, and works unchanged on a future lesson whose throughput nobody predicted.
- **The objective is unchanged per lesson.** This is a strictly monotone transform of main's existing reward function, so the optimal factory for each lesson is provably identical to main's. Only the cross-lesson gradient allocation changes: multiplicative scale gaps become additive offsets, which baselines (critic or group mean) remove for free.
- **Cost handling is untouched.** `entity_cost_scale` and `cost_efficiency` keep their current meaning and their docstring is unchanged.
- `--reward-symlog-r0 0` disables the compression; that path is main's reward exactly, and stays tested.

### Why compress the cost-adjusted reward instead of subtracting a log-space cost term

The obvious alternative, `log1p(thput_raw / r0) - log1p(entity_cost_scale * cost)`, is wrong at the bottom: it pays a **negative** reward (−0.02 to −0.2) to a factory that delivers nothing, so an empty grid outscores a nearly-complete non-delivering one — a direct incentive to build nothing. Nesting avoids that for free, because the cost multiplier is bounded in (0, 1] and `log1p` is non-negative on it.

The two forms are otherwise the same function. For `x/r0 >> 1`, `log1p(x·ce/r0) → ln(x/r0) − ln(1/ce)` — i.e. the additive penalty falls out asymptotically. Measured difference across the real (thput, cost) range:

| thput | cost | nested | additive | diff |
|---:|---:|---:|---:|---:|
| 0.000 | 242.5 | 0.0000 | −0.2171 | +0.2171 |
| 0.043 | 62.5 | 1.6188 | 1.6068 | +0.0120 |
| 0.172 | 82.0 | 2.8271 | 2.8226 | +0.0045 |
| 4.380 | 242.5 | 5.8679 | 5.8674 | +0.0006 |
| 15.000 | 20.0 | 7.2941 | 7.2941 | +0.0000 |

They agree to <0.005 for every delivering factory and diverge only where the additive form was wrong.

## Tradeoff worth noting

**A concave transform is mildly risk-averse** under stochastic outcomes (prefers a reliable 5/s over a coin flip between 0 and 15). Probably fine — arguably desirable — for factory building, but it is a real change in preferences under uncertainty.

This also does **not** create signal where reward is zero; the collapsed-assembler recovery problem is untouched.

## Tests

- `TestReward` checks whichever scheme the env is configured for, with `test_entity_cost_reduces_reward_multiplicatively_without_log` pinning main's exact behaviour at `r0=0`.
- `test_zero_throughput_cost_penalty_is_never_negative` is parametrised over `r0 ∈ {0.0, 0.01}` — the guarantee now holds under both schemes, which is the point of the nested form.
- `TestLogRewardScaleCompression` is the end-to-end guard on the point of the change: builds a solved MOVE_ONE_ITEM and a solved MEMORISE_4_INGREDIENT_RECIPES factory and asserts the reward ratio stays under 10x while the raw items/s ratio exceeds 50x.
- `test_reward_symlog_r0_default` pins the default.

Full suite: 3187 passed, 708 skipped. `ruff check` and `ty check` clean. PPO smoke test passes (the container's `torch.compile` CPU-inductor backend fails to compile on this image — reproduced identically on unmodified `main`, unrelated to this change; smoke run verified with dynamo disabled).

## Next step

Not included here, since they would confound the comparison: `norm_adv=False` and `ent_coef=0` for the finetune config. Intended validation is `/ci compare ppo --start-from <sft ckpt>` — PR (compressed) vs main (uncompressed) — with the pass/fail being whether the assembler lessons hold their SFT throughput instead of collapsing.

### DIFF
--- a/ppo.py
+++ b/ppo.py
@@ -1,4 +1,5 @@
 import json
+import math
 import os
 import contextlib
 import typing
@@ -236,6 +237,8 @@ def _run_signature(args) -> str:
         sig += f"-kl{args.target_kl:g}"
     if args.critic_warmup:
         sig += f"-cw{args.critic_warmup}"
+    if args.reward_symlog_r0:
+        sig += f"-r0{args.reward_symlog_r0:g}"
     if args.start_from:
         sig += f"-from{args.start_from}"
     sig += f"-c{layers}-seed{args.seed}"
@@ -507,11 +510,13 @@ def make_env(
     size,
     run_name,
     entity_cost_scale=PpoArgs.entity_cost_scale,
+    reward_symlog_r0=PpoArgs.reward_symlog_r0,
 ):
     def thunk():
         kwargs: dict[str, Any] = {"render_mode": "rgb_array"} if capture_video else {}
         kwargs.update({'size': size, 'max_steps': size*size, 'idx': idx,
-                       'entity_cost_scale': entity_cost_scale})
+                       'entity_cost_scale': entity_cost_scale,
+                       'reward_symlog_r0': reward_symlog_r0})
         env = gym.make(env_id, **kwargs)
         if capture_video:
             env = gym.wrappers.RecordVideo(env, f"videos/{run_name}/env_{idx}", episode_trigger=lambda e: (e+1) % 10 == 0)
@@ -577,11 +582,15 @@ class FactorioEnv(gym.Env):
         idx: Optional[int] = None,
         options: Optional[dict] = None,
         entity_cost_scale: float = PpoArgs.entity_cost_scale,
+        reward_symlog_r0: float = PpoArgs.reward_symlog_r0,
     ):
         super().__init__()
         if entity_cost_scale < 0:
             raise ValueError("entity_cost_scale must be non-negative")
+        if reward_symlog_r0 < 0:
+            raise ValueError("reward_symlog_r0 must be non-negative")
         self.entity_cost_scale = entity_cost_scale
+        self.reward_symlog_r0 = reward_symlog_r0
         if render_mode is not None:
             self.metadata = {"render_modes": [render_mode], "render_fps": 2}
             self.render_mode = render_mode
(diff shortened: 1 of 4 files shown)


beyarkay:
/ci compare ppo --start-from h76h80yb --seeds 1 --total-timesteps 1000000

PR #337, opened 2026-07-25
Fix nonzero PPO approx_kl: attention stack ignored --dropout

beyarkay:
## Root cause

`nn.TransformerEncoderLayer` defaults `dropout=0.1`, and `_SelfAttnStack` never
overrode it. With `attn_layers=4` that is **12 active `Dropout(p=0.1)` modules
plus dropout inside each of the 4 `MultiheadAttention` blocks**, regardless of
`--dropout` (which only ever reached the encoder's `Dropout2d`). `ppo.py` never
calls `.train()`/`.eval()`, so the agent runs in `nn.Module`'s default train
mode for the entire loop.

Consequence: `rollout_act` and `update_act` draw **different dropout masks**, so
the recomputed log-probs disagree with the acting ones. Every importance ratio,
`losses/clipfrac`, and `losses/approx_kl` is computed against wrong
old-log-probs — even with the actor completely frozen.

Not a `torch.compile` / CUDA-graph problem. Reproduced on CPU in eager mode:

```
act-vs-replay logprob mismatch on identical weights/inputs (eager, CPU):
  train()   approx_kl=0.115415  max|dlogp|=1.1642  mean|dlogp|=0.3483
  eval()    approx_kl=0.000000  max|dlogp|=0.0000  mean|dlogp|=0.0000
```

`eager/eager` with an `.eval()` agent was already exactly `0.0`, which is why
the existing `test_eot_logprob_roundtrips` never caught it.

### Why it was invisible until `--start-from`

`_SelfAttnStack.out_proj` is zero-init, so at fresh init the attention stage is
**exact identity** and nothing stochastic inside it can reach the outputs. A
from-scratch PPO smoke run reports `approx_kl = 0.0` even with the bug present.
It only appears once `--start-from` loads a checkpoint whose `out_proj` has
trained — exactly what run `gs1nqoni` did.

## Fix

Plumb the existing `dropout` arg into `_SelfAttnStack` →
`nn.TransformerEncoderLayer(dropout=...)`. PPO's `dropout=0.0` now makes the
whole network deterministic; nothing else changes.

Dropout has no parameters, so **checkpoint compatibility is unaffected** in both
directions.

## Behaviour change to be aware of

SFT's attention dropout now follows `SftArgs.dropout` (**0.1827**) instead of
the previously hardcoded **0.1**. SFT calls `.train()`/`.eval()` correctly, so
dropout is a legitimate regularizer there — but this does change SFT training,
and it means the arch sweeps' `dropout` axis now actually controls the attention
stage. Flagging rather than folding a second decision in: say the word and I'll
pin the attention stage to `0.1` (or `0.0`) behind its own knob instead.

Inference-only consumers (`scripts/factory_builder.py`,
`factorion-mod/server/server.py`) use the `dropout=0.0` default and call
`.eval()`, so they are unaffected either way.

## Guards

**In-loop assert** (`ppo.py`) — a CPU test cannot reach the `reduce-overhead`
CUDA-graph paths, so the invariant is also checked at runtime: during critic
warm-up the actor is frozen, so `approx_kl` must sit at the FP-noise floor
(`_WARMUP_KL_TOL = 1e-3`, vs the ~5e-4 graph-fusion budget). Fires only when PPO
is broken, on GPU as well as CPU.

**8 regression tests** (`tests/test_rl_from_checkpoint.py`), one per W&B panel
that was carrying the signal. All perturb `out_proj` so the attention stage is
non-identity — without that they have no teeth. All 8 fail before the fix, pass
after, and the whole set runs in ~10s (every forward hoisted into a
module-scoped fixture) since it only ever needs to check things at step 0.

| Test | Invariant / symptom |
|---|---|
| `test_no_dropout_is_active_at_dropout_zero` | `--dropout 0` means the whole net is deterministic, attention included |
| `test_replay_logprob_matches_sampled` | the log-prob round-trip PPO's ratio divides by |
| `test_repeated_calls_agree` | same obs+action → same logp/value (the `critic/explained_variance` symptom: stochastic features mean `V(s)` isn't a function of `s`) |
| `test_train_and_eval_mode_agree` | `train()` == `eval()` (the `eval/thput` ≫ `rollout/thput` gap at iteration 1) |
| `test_first_update_is_exactly_on_policy` | `approx_kl` and `clipfrac` are 0 on the first minibatch — zero optimizer steps have happened |
| `test_frozen_actor_keeps_kl_at_the_noise_floor` | the invariant `gs1nqoni` violated at 0.27–0.36 |
| `test_clipping_never_engages_while_frozen` | `losses/clipfrac` flat 0 across warm-up |
| `test_target_kl_does_not_trip` | `target_kl` measures policy movement, not a recompute artifact |

## Checks

Run locally, all green: `ruff`, `ty`, `cargo fmt --check`, `cargo clippy -D
warnings`, `cargo test` (163 passed), `maturin develop --release`, `pytest
tests/` (3186 passed / 708 skipped), and the PPO smoke
(`iter1_signature.approx_kl = 0.0`, all 8 epochs run, warm-up assert silent).

Two caveats, both environmental rather than code:

- The container's CPU inductor is broken independently of this change — a torch
  AVX512 codegen bug (`error: operands to '?:' have different types
  VecMask<int,1> and VecMask<float,1>`) that reproduces on unmodified `main`.
  The PPO smoke was run with `TORCHDYNAMO_DISABLE=1`; the compiled paths need CI.
- The SFT smoke was SIGTERMed by the container before finishing (resource limit,
  not a failure) — leaving that one to CI.

https://claude.ai/code/session_013VxReZNgqXw59A4LsjyFEo

---
_Generated by [Claude Code](https://claude.ai/code/session_013VxReZNgqXw59A4LsjyFEo)_

### DIFF
--- a/ppo.py
+++ b/ppo.py
@@ -111,6 +111,12 @@ _CH_FOOTPRINT = Channel.FOOTPRINT.value
 # "buildable" is exactly "not UNAVAILABLE" — only the one sentinel is named.
 _FOOTPRINT_UNAVAILABLE = Footprint.UNAVAILABLE.value
 
+# Largest approx_kl a frozen actor can legitimately produce: the rollout and the
+# update run different compiled handles at different batch sizes, so their
+# log-probs differ by ~5e-4 of graph-fusion FP ordering. Exported so the
+# regression test and the in-loop warm-up assert share one number.
+_WARMUP_KL_TOL = 1e-3
+
 
 def apply_placement_action(
     world_CWH: torch.Tensor,
@@ -1249,7 +1255,7 @@ class _SelfAttnStack(nn.Module):
     is zero-init, so the stage is identity at init and shape-preserving
     (grid-sized in == grid-sized out)."""
 
-    def __init__(self, channels, dim, heads, layers, num_tokens, pos_embed):
+    def __init__(self, channels, dim, heads, layers, num_tokens, pos_embed, dropout):
         super().__init__()
         # MultiheadAttention needs dim % heads == 0; snap heads down to the
         # largest divisor <= the requested count so any (dim, heads) the sweep
@@ -1261,10 +1267,16 @@ class _SelfAttnStack(nn.Module):
         self.pos_embed = (
             nn.Parameter(torch.zeros(1, num_tokens, dim)) if pos_embed else None
         )
+        # dropout MUST come from the caller: TransformerEncoderLayer defaults it
+        # to 0.1, and PPO cannot tolerate a stochastic forward at all — the
+        # rollout and the update would draw different masks, so the recomputed
+        # log-probs (and hence every importance ratio and approx_kl) would be
+        # wrong even with the actor frozen.
         enc_layer = nn.TransformerEncoderLayer(
             d_model=dim,
             nhead=heads,
             dim_feedforward=2 * dim,
+            dropout=dropout,
             batch_first=True,
             norm_first=True,
         )
@@ -1403,6 +1415,7 @@ class AgentCNN(nn.Module):
                 layers=attn_layers,
                 num_tokens=self.width * self.height,
                 pos_embed=bool(attn_pos_embed),
+                dropout=dropout,
             )
         num_params_encoder = sum(p.numel() for p in self.encoder.parameters())
         print(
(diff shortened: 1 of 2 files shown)


beyarkay:
/ci compare ppo --start-from h76h80yb --seeds 1 --total-timesteps 1000000

Issue #339, opened 2026-07-25
RL-from-scratch feasibility: measured exploration density, four blockers, and where the compute actually goes


beyarkay:
Feasibility study of "drop the lessons, grind reinforcement learning from scratch at scale". Everything below is measured, not estimated. Nothing here is actioned — filed as a record.

## Verdict

**Belt routing is learnable from scratch; crafting is not, and no amount of compute changes that.** A masked-random policy solves transport lessons by luck roughly 1 in 107–172 episodes, and scores 0 successes in 1,200 episodes on every assembler, splitter and underground-belt lesson. A pure from-scratch grind will learn to move items and plateau there permanently.

Separately, four small blockers currently handicap any from-scratch attempt, and the reward has a zero-point bug that plausibly explains why PPO degrades strong pretrained bases.

## Method

Masked-uniform-random policy on a fully blank 11×11 grid (source and sink placed, everything else removed), driven through the real `FactorioEnv` with the real validity rules. Per step it picks a uniform legal tile × 1 of 6 placeable entities × 1 of 4 directions; item and misc are constrained exactly as `AgentCNN.semantic_head_log_probs` constrains them (item is only a free choice for an assembler: 1 of 60 recipes). End-of-turn disabled, 121 placements per episode, throughput sampled every step.

Two caveats on the numbers: the probe always places on a legal tile (PPO's rollout does not — see blocker 3), and it never deletes (`empty` excluded from its entity set), so it cannot recover from a mistake.

## 1. Solution density — the core problem

Fully blank grid, 12,500 episodes per arm:

| policy | p(a working factory appears) | rate | throughput when it works |
|---|---|---|---|
| all 6 placeable entities | 0.70% | 1 in 142 | 0.32 of reference |
| transport belts only | 0.94% | 1 in 107 | 0.76 of reference |

Success scales with how many placements the policy gets: 20 placements → 0.16%, 60 → 0.33%, 121 → 0.70%.

Per lesson, 1,200 episodes each:

| lesson | rate | lesson | rate |
|---|---|---|---|
| MOVE_ONE_ITEM_CHAOS | 1 in 46 | MEMORISE_1..4_INGREDIENT_RECIPES | 0 in 1,200 |
| MOVE_ONE_ITEM | 1 in 172 | FACTORY_1_INGREDIENT | 0 in 1,200 |
[...]

PR #344, opened 2026-07-26
Trials: RL-only kinds with no known solution, scored apart from lessons

beyarkay:
Rebases #297 onto current `main` (it was 20 commits behind and predated the `thput_eot`→`thput` rename, `SPLITTER_MERGE_SIDELOADED`, EOT-skips-placement, action masking and the attention arch), then adds the lesson/trial metric split. #297 can be closed in favour of this.

Two self-contained commits.

## 1. Trials (`54f259e`)

A **trial** is a kind with *no known solution*: the built world is only the source/sink markers, so there is nothing to imitate and it is trained on by RL alone. `LessonKind::is_trial()` marks them, exposed as `py_lesson_is_trial()` and mirrored as `LESSON_IS_TRIAL`.

`TRIAL_RECIPE_TREE_DEPTH_1/2/3` picks a random craftable item as the sink, walks its ingredient tree randomly expanding expandable ingredients, and places one source per unique frontier item. Expansion is capped at the kind's depth and the draw is rejected unless the deepest leaf lands *exactly* on it, so each kind is a distinct rung (DEPTH_1 is exactly the direct ingredients). Markers sit on the grid edge working inward — sources face their wall's inward normal, the sink faces outward since it pulls from the cell behind it — each reserving an interior working cell. `total_entities == 0`, so blanking is a no-op and every episode starts from bare markers.

```
.. .. .. .. Sv .. .. .. .. .. ..
.. .. .. .. .. .. .. .. .. .. ..
                 (8 empty rows)
.. .. .. .. .. .. .. .. .. .. S<
.. S^ .. S^ .. .. .. .. Kv .. ..
```

Every `BuiltFactory`/`Factory` now carries `max_throughput` — the items/s reference throughput is normalized by: the solved world's simulated rate for lessons, an **analytic** single fully-fed assembler output rate for trials. `FactorioEnv.reset` reads it off the factory instead of re-simulating the solved world. First bite out of #161.

The SFT demo sampler skips trials (they can never emit an expert pair; left in the pool the fewest-pairs draw would redraw them forever). This lands in the same commit as the generator because separating them leaves an intermediate commit where the sampler hangs on the first trial it draws.

## 2. Lesson/trial metric split (`16ce472`)

Trials measure a different thing from lessons — inventing a factory versus reproducing one the generator can also build — so pooling them would silently move the headline number the SFT baseline and every historical run are quoted in. SFT never evaluates a trial, so a mixed `eval/thput` would no longer overlay on `val/thput`.

- `run_rollout_eval` keeps separate pools: `overall`/`overall_eot` stay **lesson-only**; `trial_overall`/`trial_overall_eot`/`trial_n` carry the trial side.
- PPO logs `eval/trial_thput`, and a trial episode's rollout aggregates land under `rollout/trial_*` instead of `rollout/*` — the whole block, so each side describes one population rather than a reward mean and a throughput mean taken over different sets.
- The moving average behind the pbar and the final run tag follow the same split.
- Per-kind `{eval,rollout}/{TRIAL_*}/…` keys are already name-separated and are logged as usual, so each trial's own curve stays visible.

`eval/trial_thput` is the point of the PR: it scores building a factory from nothing but the markers, which is the actual problem — every lesson is a proxy for it. Expect it to sit at zero for a long time; per #339 a masked-random policy scores 0 successes in 1,200 episodes on every crafting lesson, and trials are strictly harder.

## Tests

- Rust: `test_recipe_tree_trial_smoke` (markers-only worlds, frontier validity, exact-depth chains, analytic ceiling, edge/orientation + working-cell invariants); the no-orphan test skips trials, whose markers are unconnected by construction.
- Python `tests/test_recipe_tree_trial.py`: trial flags, blank no-op, env episode starts from markers, SFT stream never draws a trial, frontier validity, edge/orientation invariants — plus `TestTrialEvalIsScoredApart` (adding trials to an eval set must not move `eval/thput`; trials report under their own keys; `_run_greedy_eval` surfaces `eval/trial_thput`).
- `tests/test_ppo_logging.py::TestTrialRolloutMetricsAreSeparate`: nothing from a trial reaches the lesson aggregate, both sides carry the same field names, per-kind keys unaffected.

## Checklist

`cargo fmt`/`clippy -D warnings` clean, 164 Rust tests, `maturin develop --release`, 3224 Python tests passed / 708 skipped, `ruff` clean, `ty` clean, both smoke tests exit 0.

One caveat: the PPO smoke test needs `TORCHDYNAMO_DISABLE=1` **in this container** — torch's CPU inductor backend fails C++ codegen (`VecMask<int,1>` vs `VecMask<float,1>`) on the attention stack. That reproduces independently of this diff and does not affect CUDA CI pods.

## Not done (flagging, not folding in)

The global `critic/*` diagnostics still pool trials with lessons. That seemed right — the value head is one head trained on all transitions — but say the word if you want it split too. Per-lesson `critic/{TRIAL_*}/…` already breaks out.

https://claude.ai/code/session_01Pv8PGDzBRYsWQR84S7vhtE

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pv8PGDzBRYsWQR84S7vhtE)_


beyarkay:
/ci ppo --start-from h76h80yb --total-timesteps 1000000 --track --clip_coef=0.14110262908153953 --dropout=0 --ent_coef_end=0 --ent_coef_start=0.009194859414592384 --entity_cost_scale=1.5432189370835023e-05 --gae_lambda=0.9579809448126424 --gamma=0.9895902232063256 --learning_rate=2.4328405139207804e-05 --max_grad_norm=1.8089018007290625 --num_minibatches=16 --reward_symlog_r0=0.012829079584619752 --target_kl=0.017221367557920843 --update_epochs=8 --vf_coef=1.2791352970926626

(args coming form [this well performing run](https://wandb.ai/beyarkay/factorion/runs/um2169cm/overview?nw=nwuserbeyarkay))

beyarkay:
/ci compare ppo --start-from h76h80yb --total-timesteps 1000000

PR #260, opened 2026-07-06
Focused PPO sweep: vary step_penalty to rescue EOT-head collapse

beyarkay:
## Summary

Replace the broad Bayesian hyperparameter sweep with a focused grid sweep over `--step-penalty` values `[0, 0.0025, 0.005, 0.0075, 0.01]` to investigate whether per-step length penalties can rescue the EOT-head collapse observed in run y324x0d4.

## Key Changes

- **Sweep method**: Changed from `bayes` (30 runs, 11 hyperparameters) to `grid` (5 runs, 1 hyperparameter)
- **Metric**: Kept `eval/thput_eot` (maximize) as the target
- **Baseline**: Finetune from SFT base `uju9n8ql` (the collapsed base) with `--start-from uju9n8ql` to directly measure whether step penalties fix the collapse
- **Hyperparameter isolation**: All other hyperparameters (lr 7e-4, ent 0.02041→0.0004576, kl 0.02, cw5, layers 93/69/96) held at `training_config.py` defaults, reproducing the collapsed run at `step_penalty=0`
- **Diagnostic goal**: Measure whether the collapse (eot_rate 0.87 → 0, length → 123 truncation, thput → 0) is rescued by varying step penalties

## Implementation Details

- Reduced `run_cap` from 30 to 5 (one per grid point)
- Added `--start-from uju9n8ql` to the command template so all runs finetune from the same collapsed SFT base
- Kept `--total-timesteps 250_000` and `--track` flags
- Preserved sweep infrastructure (method, metric, program, command structure)

https://claude.ai/code/session_01Hcpv1KF7MY5CF9us5o39xF

PR #346, opened 2026-07-30
Fold sweep llfgwhki's cross-metric consensus into PPO defaults

beyarkay:
## Why

[Sweep `llfgwhki`](https://wandb.ai/beyarkay/factorion/sweeps/llfgwhki) (48 bayes runs, metric `eval/thput`, `--start-from h76h80yb` at 1M steps, seed 1, today's arch) was **never folded into the defaults**. Every PPO value in `training_config.py` still traced to `6d1768b` (#260, 2026-07-07) — nineteen days before the sweep. The sweep's own 14-param config also exists in no commit; `ci/sweep_ppo.yaml` on `main` is a different (4-value `entity_cost_scale` grid) config, so `llfgwhki` was launched ad-hoc.

## What changed

Re-ranking the same 48 runs by `rollout/thput` instead of `eval/thput` agrees only at ρ=+0.51 and reshuffles the winner (`wnerxtpb` 1st → 3rd; `um2169cm` eval-rank 7 → rollout 1st). Rather than pick a metric, this uses the second ranking as a **filter**: adopt only axes where both metrics agree in direction, taking the eval-winner's value.

| Param | was | now | ρ_eval / ρ_rollout |
|---|---|---|---|
| `gamma` | 0.9566 | 0.9829 | +0.11 / −0.14 (both winners ≈0.98) |
| `gae_lambda` | 0.9187 | 0.9832 | +0.10 / +0.25 |
| `num_minibatches` | 32 | 8 | −0.14 / −0.40 |
| `max_grad_norm` | 1.221 | 1.9171 | +0.26 / +0.34 |
| `target_kl` | 0.02 | 0.0089354 | −0.29 / −0.27 |

`learning_rate` (3.369e-05) and `reward_symlog_r0` (0.01) are in the agreeing group too, but already sit where both metrics point — no change needed. `gamma`/`gae_lambda` are precisely the axes `EXPERIMENT_LOG.md` flagged as inherited `j0s5y2mc` values that were never independently validated.

**Deliberately left alone**, because the cross-metric check says the eval winner's draws there aren't evidence:

- `clip_coef` — ρ flips sign, +0.09 → **−0.40** (eval#1 0.262 vs rollout#1 0.141).
- `entity_cost_scale`, `vf_coef`, `ent_coef_start` — signal vanishes on rollout (+0.16/+0.16/+0.18 → +0.01/+0.03/+0.02).
- `update_epochs` — weak/noise both ways.
- `ent_coef_end` — pinned at 0 for the entire sweep, so it says nothing about today's 0.0007372.

## Caveats, stated plainly

1. **Single-seed, narrow band.** All 48 runs land in `eval/thput` 0.678–0.807 (median 0.742); rank 1 beats rank 10 by 0.04. This fold is a hypothesis, not a measured win. It wants `/ci compare ppo --start-from h76h80yb --seeds 3` with `assert pr:eval/thput > main:eval/thput` before anyone trusts it.
2. **`target_kl` overrides a 5-seed result.** `EXPERIMENT_LOG.md:529` records 0.02 as part of a 5-seed-confirmed time-to-quality recipe. Its rationale — early-stopping the epoch loop to make a high LR cheap per-iter — assumed that recipe's `lr 7e-4`, which is 20× today's 3.369e-05, so the stated reasoning no longer applies as written. The docstring now cites both sources. Worth a maintainer's call; reverting this one line is easy if you disagree.
3. `tests/test_rl_from_checkpoint.py::test_target_kl_does_not_trip` reads `PpoArgs().target_kl` and asserts warm-up KLs stay under it, so the lower ceiling tightens that assertion.

## Testing

- `uv run ruff check .` — pass
- `uv run ty check .` — pass
- **PPO smoke test — cannot run in this container, and this is pre-existing, not caused by the change.** `torch._inductor` fails to compile its own generated CPU C++ for `rollout_act` (`error: operands to ?: have different types VecMask<int,1> and VecMask<float,1>`). Reproduced identically with the *pre-change* values passed explicitly on the CLI (`--gamma 0.9566 --gae-lambda 0.9187 --num-minibatches 32 --max-grad-norm 1.221 --target-kl 0.02`) → same `CppCompileError`, same exit 1. It's a CPU-only box; the CUDA CI pods should be unaffected.
- **`pytest` and an eager-mode (`TORCHDYNAMO_DISABLE=1`) smoke run were skipped at the author's request** — CI should be the gate on those, in particular the `target_kl` test named in caveat 3.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W6uNQxzBQYLDj6GqnoHyBb)_

### DIFF
--- a/training_config.py
+++ b/training_config.py
@@ -121,11 +121,11 @@ class PpoArgs(SharedArgs):
     """the number of steps to run in each environment per policy rollout"""
     anneal_lr: bool = True
     """Toggle learning rate annealing for policy and value networks"""
-    gamma: float = 0.9566
+    gamma: float = 0.9829
     """the discount factor gamma"""
-    gae_lambda: float = 0.9187
+    gae_lambda: float = 0.9832
     """the lambda for the general advantage estimation"""
-    num_minibatches: int = 32
+    num_minibatches: int = 8
     """the number of mini-batches. more minibatches -> smaller minibatch size -> more likely to fit on GPU"""
     update_epochs: int = 8
     """the K epochs to update the policy"""
@@ -164,9 +164,9 @@ class PpoArgs(SharedArgs):
     opposed to subtracting a log-space cost term) also keeps zero throughput at
     exactly zero, so a factory that delivers nothing is never worse than an
     empty grid. 0 disables the compression."""
-    max_grad_norm: float = 1.221
+    max_grad_norm: float = 1.9171
     """the maximum norm for the gradient clipping"""
-    target_kl: Optional[float] = 0.02
+    target_kl: Optional[float] = 0.0089354
     """the target KL divergence threshold; early-stops the update's epochs. None
     = always run all update_epochs. (Why this default: EXPERIMENT_LOG.md.)"""
     adam_epsilon: float = 6.891e-06


beyarkay:
/ci compare ppo --start-from h76h80yb --total-timesteps 2500000 --seeds 3

beyarkay:
Not sure why the bot didn't add all 6 runs to the statistical comparison. manually ran it via claude:



> 
> metric | main | pr | Δ | p
> -- | -- | -- | -- | --
> eval/thput | 0.7758 | 0.7836 | +0.0078 | 0.376
> eval/trial_thput | 0.0370 | 0.0410 | +0.0040 | 0.423
> rollout/thput | 0.7880 | 0.7934 | +0.0054 | 0.289
> rollout/reward | 4.4609 | 4.4462 | −0.0147 | 0.845
> 
> 
> 
> <p><strong>What did move significantly</strong> (33 of 246 at p&lt;0.05, vs ~12 expected by chance):</p>
> <ul>
> <li><strong>Policy entropy up</strong> — <code>policy/entropy_tile</code> 0.659 → 1.095 (p=0.0009), <code>policy/entropy</code> 0.917 → 1.365 (p=0.0027). The PR policy stays markedly less peaked, which is what <code>target_kl</code> 0.02 → 0.0089 should do: earlier epoch-loop early-stopping means less per-iteration sharpening.</li>
> <li><strong>Critic scale up</strong> — <code>critic/return_mean</code> 3.32 → 4.01 (p=0.0001), <code>value_mean</code> +0.68. That's mechanical from <code>gamma</code> 0.9566 → 0.9829, not a quality signal.</li>
> <li><strong>~16% slower</strong> — <code>perf/sps</code> 279 → 235 (p=0.0103). But the cause is <code>perf/eval_seconds</code> 14.6 → 19.2 (p=0.0006), while <code>perf/update_seconds</code> is unchanged (16.54 vs 16.62). So <code>num_minibatches=8</code> costs nothing per update; the slowdown is more env steps from the higher-entropy policy running longer episodes (<code>rollout/length</code> 10.23 → 10.70).</li>
> <li>Several <code>critic/*/explained_variance</code> and <code>value_rmse</code> rows degrade (<code>MOVE_VIA_UG_BELT</code> EV 0.618 → 0.438, <code>SPLITTER_MERGE_SIDELOADED</code> 0.405 → 0.210) — expected, since the critic is now fitting a larger-magnitude return target.</li>
> </ul>
>

PR #347, opened 2026-07-30
Skip the reduce-overhead compile on CPU so CPU runs work

beyarkay:
## The bug

Every CPU PPO run dies before the first training step — including the CPU-only PPO smoke test from the pre-completion checklist:

```
torch._inductor.exc.InductorError: CppCompileError: C++ compile error
...main.cpp:142:39: error: operands to '?:' have different types
  'at::vec::CPU_CAPABILITY::VecMask<int, 1>' and 'at::vec::CPU_CAPABILITY::VecMask<float, 1>'
```

Inductor's CPU vectoriser miscodegens the `_legal_tile_mask` select. The generated kernel builds an int32 mask in a lambda but pairs it with a float mask in the else arm of one ternary:

```cpp
auto tmp13 = ~tmp12;                    // VecMask<int32_t,1>
return tmp13;
...
auto tmp14 = tmp2 ? tmp3() : at::vec::VecMask<float,1>::from(static_cast<bool>(0));
```

g++ rejects it, and the `CppCompileError` surfaces as an `InductorError` out of the first `rollout_act()` call — so the run never reaches a rollout step. This is an upstream inductor bug (torch 2.12.1), triggered by our mask rather than caused by our code.

## The fix

The three inference handles now go through a one-line `maybe_compile(fn, device)` that passes the callable through untouched **on CPU only**.

MPS deliberately keeps compiling. The bug is in inductor's C++ backend (`cpp_builder` → g++), which MPS never reaches — it lowers through the Metal backend (`mps` is in inductor's `GPU_TYPES`) and still gets fusion. All it loses is CUDA-graph replay, which inductor already gates on a lone `cuda` device (`cudagraph_utils.py:283`), so `reduce-overhead` was silently degrading to plain `torch.compile` on MPS regardless. Excluding MPS would have disabled a working optimization on Mac dev for a bug it can't hit.

+14/−4 in `ppo.py`, with none of the existing rationale comments touched.

## Tests

New `tests/test_compile_inference_paths.py`:

- CPU returns the callable as-is, with no dynamo wrapper, so inductor is never invoked
- `cuda` and `mps` still get a compiled wrapper — `torch.compile` is lazy, so this pins the device gate without needing either accelerator present
- the passed-through CPU callables really run the policy (action + value of the expected shape)

## Verification

Passing: `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test`, `ruff check .`, `ty check .`, and the new test module (4 passed).

The full `pytest tests/` and both smoke tests are running against this commit; results to follow in a comment. An earlier PPO smoke run against the same CPU behaviour exited 0, with the run stepping (`taking step 10/256`) where it previously crashed.

Note that `tests/test_torch_compile.py` never caught this: it calls `torch.compile(agent)` in default mode, which — per the long-standing comment above the callsite — proxies `.get_action_and_value` straight to eager and compiles nothing.

### DIFF
--- a/ppo.py
+++ b/ppo.py
@@ -1206,6 +1206,16 @@ def assert_device_ok(device) -> None:
     )
 
 
+def maybe_compile(fn, device):
+    """`torch.compile(reduce-overhead)` on everything but CPU, where inductor's
+    vectoriser miscodegens the `_legal_tile_mask` select (`VecMask<int,1>` vs
+    `VecMask<float,1>` in one ternary) and g++ rejects the kernel before the
+    first rollout step. MPS keeps compiling: it reaches the Metal backend, not
+    the C++ one, and still fuses — it just skips the CUDA-graph replay, which
+    inductor only ever enables for a lone cuda device."""
+    return fn if device.type == "cpu" else torch.compile(fn, mode="reduce-overhead")
+
+
 def cuda_env_info() -> dict[str, str]:
     """Wheel CUDA build + host driver, recorded in every run's W&B config.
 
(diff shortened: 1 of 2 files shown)


beyarkay:
## Pre-completion checklist: all green on `30c399d`

| Check | Result |
|---|---|
| `cargo fmt --check` | pass |
| `cargo clippy -- -D warnings` | pass |
| `cargo test` | pass |
| `maturin develop --release` | built at session start; no Rust changes in this PR |
| `pytest tests/` | exit 0 — **3234 passed, 708 skipped** (37m11s) |
| `ruff check .` | pass |
| `ty check .` | pass |
| PPO smoke (`--total-timesteps 5000`) | exit 0 — 2m08s |
| SFT smoke (`--size 5 --num-samples 200`) | exit 0 |

The PPO smoke log confirms the gate and the absence of the failure:

```
Inference paths on cpu: eager
...
rollout=11.4s (steady mean 11.41s/iter), update=72.7s (steady mean 72.67s/iter) over 1 iters
```

Zero occurrences of `Traceback`, `InductorError`, or `CppCompileError` — where before the fix the run died at the first `rollout_act()` call. Both the eager rollout path and the eager update path execute (the run completes its 8 optimiser epochs).

Earlier attempts at these two smoke tests timed out at exit 124. That was contention, not the code: this container has 4 cores and I had it at load average ~11 running three jobs at once. Run serially with the box to itself, the PPO smoke takes 2m08s.

---

PR #352, opened 2026-07-30
Update training_config.py

beyarkay:

PR #353, opened 2026-07-31
Reward closing the source→sink gap, scaled by each lesson's ceiling

beyarkay:
## Why

`calc_throughput` keeps only `F ∩ B` — the intersection of the forward-reachable set from sources and the backward-reachable set from sinks — so an entity earns credit only once the chain closes end to end. A half-built factory is indistinguishable from an empty grid.

That means lessons the policy can't finish (`TRIAL_RECIPE_TREE_DEPTH_*`, `FACTORY_1_INGREDIENT`) have a **perfectly flat objective**: doing nothing and vandalising the whole grid both return exactly 0. `gamma < 1` was accidentally supplying the only gradient there, and it pointed at "fire EOT immediately" — which is what collapsed trial episode length to ~2.7 steps in [io59q06n](https://wandb.ai/beyarkay/factorion/runs/io59q06n). Removing the discount (#352) fixed the collapse but replaced it with the opposite failure: `trial_invalid_frac` 0.003 → 0.488, entropy rising instead of annealing, because nothing at all opposed flailing.

## What

A dense terminal term scoring how much of the source→sink gap has been closed:

```
reward += connect_coef * almost_connected_reward * log1p(max_throughput / reward_symlog_r0)
```

`almost_connected_reward` (Rust, reusing the two reachability sets that already exist) is, per source, the fraction of its initial distance to the nearest sink that the connected structure has eaten — counting growth from either end, averaged over sources so wiring 2 of a trial's 3 ingredient lines scores 2/3.

**Measuring the residual gap rather than chain length** is what makes it unfarmable: a belt spiral that never approaches the sink earns nothing (covered by a test).

**Scaling by each lesson's own ceiling** is load-bearing. Achievable rates span **1600×** — a solved `MEMORISE_4` runs at 0.00172 items/s and pays only **0.159** total reward, while belt lessons pay 7.31. Any flat coefficient large enough to matter on a belt lesson would let a half-built grid outrank a *perfect solve* on a slow-recipe one. Scaled, the bonus is 25% of a solve on all 14 lessons alike. `max_throughput` is the generator's per-episode ceiling set in `reset()`, not anything the agent built, so that scale is fixed and positive for the whole episode.

The effect, on `MOVE_ONE_ITEM` with entities removed from the solved factory:

| `num_missing_entities` | `thput_raw` | reward |
|---|---|---|
| 0 (solved) | 15.000 | 9.12 |
| 1 | **0.000** | **1.31** |
| 2 | **0.000** | **0.52** |
| 3 | **0.000** | **0.26** |
| ∞ (blank) | 0.000 | 0.00 |

The three middle rows deliver nothing — on main they all pay exactly 0.0 and are indistinguishable. That ladder is the feature.

## Notes

- `run_rollout_eval` scores `info['thput_normed']` and never the reward, so **`eval/thput` and `eval/trial_thput` are unaffected** and stay directly comparable to main and the SFT baseline. Gate any compare on those, not on `rollout/reward` (whose scale this changes).
- Other sources land in the backward set (they reach the sink), but nothing can be attached to a source, so they're excluded as gap targets — otherwise an unwired source earns credit for sitting near a wired one.
- `connect_coef` is on `SharedArgs`; `SftArgs` inherits it but SFT has no reward, so it's unused there. `connect_coef = 0` disables the shaping.
- Logged as `rollout/almost_connected_reward`, `rollout/trial_almost_connected_reward`, and `rollout/{LESSON}/almost_connected_reward`.

## Cost

The gap scan is `|forward| × |backward|`, and both sets grow with every connected tile placed, so it is quadratic in grid area. Worst case is two snakes each covering half the grid — one source-fed, one sink-feeding, never touching (release build, set sizes verified by BFS on the real engine graph):

| grid | tiles | \|F\| | \|B\| | µs/call |
|---|---|---|---|---|
| 11×11 | 121 | 89 | 109 | 144 |
| 20×20 | 400 | 361 | 359 | 734 |
| 32×32 | 1024 | 961 | 959 | 2,190 |
| 50×50 | 2500 | 2401 | 2301 | 7,244 |
| 120×120 | 14400 | 14161 | 14159 | 138,550 |

For reference `simulate_throughput` is 42 µs on a real factory, and this runs under the same terminal-step-only guard. Fine at the current 11×11 and at 20×20; it would dominate the rollout well beyond that. Note a *flailing* agent does not trigger it — dense random fill measures roughly linear, because belt connectivity fragments both sets. It takes a competent agent that has built most of a factory without closing it.

## Testing

Full pre-completion checklist, green on this tree (rebased onto b8f8b0c):

| check | result |
|---|---|
| `cargo fmt --check` | clean |
| `cargo clippy -- -D warnings` | clean |
| `cargo test` | 173 passed |
| `maturin develop --release` | built |
| `pytest tests/` | 3252 passed, 708 skipped, 0 failed |
| `ruff check .` | clean |
| `ty check .` | clean |
| PPO smoke (5k timesteps) | exit 0 |
| SFT smoke | exit 0 |

- 7 Rust unit tests: markers-only is 0, monotone growth toward the sink, saturation at 1 on connection, spiral-away earns nothing, backward-from-sink counts, multi-source averaging, no-sink is 0, plus the packed-grid stress case above. That one guards its own stress — it asserts the graph stays large and the score just under 1, so a connectivity change that fragments the snakes can't silently turn it into a no-op.
- 4 Python end-to-end tests asserting rewards the env actually paid, with no formula recomputed in the test: the zero-throughput ladder above, solved beats partial, the max bonus stays under a real solve on the slowest lesson, and `connect_coef=0` pays exactly the unshaped reward.

## Validating that this actually helps

Not yet run. Notes for whoever does:

- **Gate on `eval/thput` and `eval/trial_thput`, never `rollout/reward`** — this change adds a term to the reward, so the PR side wins by construction there and it measures nothing.
- **Three-way, not two-way**: main vs `gamma=1` alone (#352) vs `gamma=1` + this. Both endpoints are already known to fail differently (quit-instantly vs flail), so a two-way can't separate "shaping helped" from "the gamma change helped".
- **Match the anneal schedule to io59q06n.** The #352 compare ran 1.5M fully-annealed and did not reproduce the collapse at all — its main side's `FACTORY_1_INGREDIENT` *improved* to 0.240. io59q06n is a 15M run observed at 1.36M, where `ent_coef` was still ~0.0075.
- **Mechanism check**: `rollout/trial_almost_connected_reward` should climb. If it is flat, the policy is not responding to the new gradient and nothing else in the run matters.
- **Known local optimum to rule out**: the score reaches 1 on a graph that closes with the *wrong* assembler recipe — connected, delivering nothing, collecting the full 25%. If that metric approaches 1.0 while `eval/trial_thput` stays at 0, that is what happened; `eval/asm_item_acc` is the tell.
- `connect_coef = 0.25` is a first guess, not a tuned value. Worth sweeping (0.1 / 0.25 / 0.5).

## Related

- #355 — the engine's assembler ignores `crafting_time`, so trial `max_throughput` is the recipe's per-craft quantity (1.0 for 93% of trials). Consumed as-is here; that issue records the problem separately.

https://claude.ai/code/session_01KnwrmLYEKEwqXqmpx11nwP


beyarkay:
/ci compare ppo --start-from h76h80yb --seeds 3 --total-timesteps 1500000
assert pr:eval/thput >= main:eval/thput

beyarkay:
Doesn't look great, maybe we sweep the coefficient but I'm not hopeful

beyarkay:
## Result: this did not work. Post-mortem for anyone picking it up later

The compare at `b23dad8` came back with **both assertions green**. Both were wrong. Recording what happened and why, so nobody re-runs this shape of experiment.

### 1. The green assertions were an artifact — do not trust `assert` on these metrics

The report said `pr:eval/trial_thput=0.03604 >= main:eval/trial_thput=0.03265`. But `eval/trial_thput` is declared `summary="max"` in `ppo.py`'s `define_metric` block, so the assertion compared the **maximum over training**, not the final value. On a metric that lives near zero and is noisy, that reduces to "which run got the single luckiest eval".

Gaussian-smoothed (σ=12 iterations), averaged over 3 seeds, the real trajectory is:

| `eval/trial_thput` | 15% | 45% | 75% | end |
|---|---|---|---|---|
| main | 0.0171 | 0.0184 | 0.0194 | **0.0179** |
| pr | 0.0154 | 0.0081 | 0.0042 | **0.0030** |

The PR **monotonically destroys** the metric it was built to improve — 6× worse, decaying the whole way. Reported as +0.0034; actually −0.0150.

Same for the target lesson:

| `rollout/FACTORY_1_INGREDIENT/thput` | 15% | 45% | 75% | end |
|---|---|---|---|---|
| main | 0.1399 | 0.0985 | 0.0796 | **0.0902** |
| pr | 0.1025 | 0.0324 | 0.0375 | **0.0282** |

**Lesson: any `assert` against a `summary="max"` metric is a coin flip. Smooth and compare final values.**

### 2. The change was tested in a broken regime (`gamma=1.0`)

This branch carried `gamma=1.0` from #352. Sweep [v3ohvfpl](https://wandb.ai/beyarkay/factorion/sweeps/v3ohvfpl) shows that is the wrong side of a cliff:

| gamma | `policy/entropy` (last) | `invalid_frac` (last) |
|---|---|---|
| 0.9515 | 1.09 | 0.013 |
| 0.9788 | 1.21 | 0.028 |
| 0.9836 | 1.10 | 0.017 |
| **0.9916** | **3.97** | **0.077** |

Above gamma ≈ 0.99 the policy stops annealing and diffuses. In this compare that showed up as `policy/entropy` 3.18 → 4.85 → 3.98 (rising, where main anneals 2.05 → 1.11) and `rollout/trial_invalid_frac` 0.126 → 0.417 (main: 0.021 → 0.0003). Those figures are near-identical to the gamma=1-only run in #352 (entropy 2.66 → 4.09, invalid 0.065 → 0.49), i.e. **the shaping was roughly neutral against the gamma pathology** — gamma did the damage.

**Lesson: keep gamma near 0.96. Do not re-test gamma → 1; it is already measured and it breaks the entropy schedule.**

### 3. The shaping was implemented terminal-only, which defeats the point

This is the substantive design error. Potential-based shaping telescopes to `Φ(s_T) − Φ(s_0)`, so it was implemented as a single terminal bonus. That preserves *total return* and throws away the only thing dense shaping is for: **per-step credit assignment**.

With GAE at `gamma*lambda = 0.9566 * 0.9187 = 0.879`, a step 30 before the end receives `0.879^30 ≈ 2%` of the terminal signal. So the "dense" reward was not dense — it was a differently-shaped terminal reward, discounted away exactly like the throughput term it was meant to counteract.

The evidence fits: the mechanism *did* engage — `rollout/trial_almost_connected_reward` reached 0.53 and `rollout/FACTORY_1_INGREDIENT/almost_connected_reward` 0.55 — and bought nothing, because the agent never learns *which placement* earned it.

**Lesson: if you want dense shaping, apply it per step (`F_t = gamma*Φ(s_{t+1}) − Φ(s_t)`), not collapsed to the terminal step. The telescoping identity is about return, not about advantages.**

### 4. Sweeping `connect_coef` would not have rescued it

`norm_adv=True` (`ppo.py:2374`) normalizes advantages to zero mean / unit variance per minibatch, so scaling the terminal reward is washed out before it reaches `pg_loss`. `connect_coef` sets the **relative** weight of connect-vs-deliver; it cannot shift the balance between the policy gradient and the entropy bonus, which is what was actually broken.

**Lesson: tuning a coefficient cannot fix a credit-assignment problem, and under `norm_adv` a global reward scale is close to a no-op.**

### What survived

- `eval/thput` was flat-to-slightly-up (0.7292 → 0.7323), so the shaping did **not** damage the lessons that already work. The failure is confined to trials, `FACTORY_1_INGREDIENT`, and on-policy sampled behaviour.
- `almost_connected_reward` itself (the Rust signal) is sound and cheap, and measures what it claims. Its cost is quadratic in grid area (see the PR body table) but negligible at 11x11.
- The underlying diagnosis still stands: throughput is 0 for every partially-built factory alike, so trials have a flat objective. That problem is real and unsolved; only this particular remedy failed.

### Suggested next attempt

Per-step potential-based shaping at gamma ≈ 0.96, which is the combination that was never tested — per-step signal is the only form that can oppose the discount's pull toward firing EOT early, since it pays at the moment progress happens rather than at an end that gets discounted away. Costs ~+11% rollout time (23.6us/step against ~210us/step) because it breaks the terminal-only fast path at `ppo.py:885`. Sweep `connect_coef` only after that, in a regime where it means something.

---

PR #363, opened 2026-08-02
Sweep gamma

beyarkay:


beyarkay:
Gamma didn't seem to have a better value, the 0.99 seemed to be mostly ranodom luck an d0.96 (existing) was also pretty good, closing

PR #365, opened 2026-08-02
Drop the autoreset junk transition from the PPO update (#233)

### LINKED ISSUE #233: PPO trains on one garbage transition per episode (Gymnasium NEXT_STEP autoreset)
## Problem

Gymnasium 1.x vector envs use NEXT_STEP autoreset: when a sub-env's episode ends at `step()` call N, call N+1 **ignores the action**, resets the env, and returns the reset obs with `reward=0, terminated=False`. Verified empirically against the installed gymnasium 1.3.0:

```
call 1: obs=[2]   r=[1.] term=[True]    <- episode ends
call 2: obs=[100] r=[0.] term=[False]   <- action IGNORED, reward forced to 0
```

The PPO rollout loop stores that call-N+1 transition like any other: `obs` = the previous episode's **terminal** observation, `action` = a sampled action the env never executed, `reward` = 0. The update then trains on it with no masking. Per episode end this means:

- The policy gradient reinforces/punishes an action that was **never executed**, with advantage `γ·V(new episode's first obs) − V(terminal obs)` — bootstrapping **across the episode boundary** into an unrelated fresh factory.
- The value head is trained to predict `γ·V(s₀')` for terminal states instead of ~0; terminal states look like ordinary mid-build states, so the error generalizes.
- The junk advantages pollute per-minibatch advantage normalization (`norm_adv`).

The poisoned fraction of every batch is `1 / mean_episode_length` — under 1% at full 123-step episodes, but 10%+ if the EOT head fires early (which the step-penalty issue actively encourages).

## Fix

The junk steps are exactly the flattened batch indices where `dones == 1` (`dones_SE[step] = next_done` is stored *before* the env step, so the flag already marks them). Exclude those indices from the policy / value / entropy losses in the update loop.

GAE needs no change: the preceding real step is already cut off correctly by `nextnonterminal = 0`, so the contamination is confined to the junk step itself.

This is also what Gymnasium's own migration guide recommends for CleanRL-style loops under NEXT_STEP autoreset.

beyarkay:
Fixes #233. One commit, `ppo.py` + a new test file.

## The problem

Gymnasium's NEXT_STEP autoreset makes the `step()` call right after an episode ends ignore the action, reset the sub-env and force `reward=0`. The CleanRL-style rollout stored it like any other transition, so every episode contributed one row whose action was never executed and whose "next state" is an unrelated fresh factory. At the current mean episode length (~10.8) that's ~8-9% of every batch:

- the policy gradient reinforced or punished an action the env never ran, with an advantage that bootstraps **across the episode boundary** into a different factory;
- the value head was trained to predict `γ·V(next factory's blank state)` at terminal states, which look like ordinary mid-build states, so the error generalises;
- the junk advantages polluted per-minibatch advantage normalisation.

## The fix

`dones_SE[step] = next_done` is stored *before* the env step, so it already flags exactly those rows. They now carry zero weight in the policy, value and entropy losses, in the advantage normalisation, in the approx-KL/clipfrac diagnostics and in the `critic/*` report, via a new `_masked_mean` helper.

GAE is unchanged, as the issue notes: the preceding real terminal step is already cut off by `nextnonterminal = 0`, so the contamination was confined to the junk step itself.

## Tests

`tests/test_autoreset_masking.py`:

- the autoreset step really does drop the action and pay 0 (pinning the Gymnasium behaviour the fix depends on);
- the stored `dones` flag marks exactly the zero-reward junk rows;
- `_masked_mean` ignores masked rows, returns 0 rather than NaN for an all-masked minibatch, and propagates **no gradient** to a masked row.

Full checklist green: `cargo fmt`/`clippy`/`test` (166), `maturin develop --release`, pytest (3253 passed), `ruff`, `ty`, and both smoke tests.

## Not in this PR

The two compounding problems from the same report are left for follow-ups: the training rollout not legal-masking the tile head (#339 blocker 3) and CROSS_UNDER_BELT paying a free 0.25 at step 0 (#339 blocker 1, both the deletable-protected-tile landmine and the reward baseline).

https://claude.ai/code/session_01ASW26Mqyqpq2hFEX4Pkh2y

### DIFF
--- a/ppo.py
+++ b/ppo.py
@@ -415,6 +415,16 @@ def _critic_stats(values: np.ndarray, returns: np.ndarray, advantages: np.ndarra
     }
 
 
+def _masked_mean(x_B: torch.Tensor, valid_B: torch.Tensor) -> torch.Tensor:
+    """Mean of ``x_B`` over the entries ``valid_B`` marks 1, ignoring the rest.
+
+    The floor of 1 on the denominator keeps an all-masked minibatch at 0
+    instead of NaN — every term is zero-weighted there anyway, so it
+    contributes no gradient.
+    """
+    return (x_B * valid_B).sum() / valid_B.sum().clamp(min=1.0)
+
+
 def _critic_diagnostics(
     values: np.ndarray,
     returns: np.ndarray,
@@ -2300,6 +2310,15 @@ if __name__ == "__main__":
         advantages_B = advantages_SE.reshape(-1)
         returns_B = returns_SE.reshape(-1)
         values_B = values_SE.reshape(-1)
+        # Gymnasium's NEXT_STEP autoreset (#233): the call right after an
+        # episode ends ignores the action, resets the sub-env and forces
+        # reward=0. Its stored transition is therefore junk — the action was
+        # never executed and its "next state" is an unrelated fresh factory —
+        # and `dones_SE[step] = next_done` already flags exactly those steps.
+        # Zero-weight them everywhere the update reads the batch. GAE needs no
+        # such guard: the preceding real terminal step is already cut off by
+        # `nextnonterminal = 0`, so the junk is confined to the junk step.
+        valid_B = 1.0 - dones_SE.reshape(-1)
 
         # Optimizing the policy and value network
         update_start = time.time()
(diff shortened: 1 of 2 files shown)


beyarkay:
/ci compare ppo --start-from h76h80yb --seeds 3 --total-timesteps 2500000

beyarkay:
Soo this is a bugfix, and it mostly performs worse and the per-lesson thput, *except* on the lessons we actually care about (single-depth trials and FACTORY_1_ITEM) so it's being merged.

Issue #371, opened 2026-08-05
PPO research loop: working guide, do-not-retest list, and experiment log


beyarkay:
Durable working guide for the autonomous PPO-improvement loop. Goal: **`eval/trial_thput` up** (primary), `eval/thput` no regression (guardrail).

**Current state (2026-08-10): #374 MERGED by owner. Lazy-equilibrium mechanism now env-verified (see below). Nothing running. Pre-merge "main" baselines are stale for the new main tip.**

## The loop (per experiment)
1. Highest expected information per unit time; de-risk with cheap local probes before pods.
2. Hypothesis + falsifiable predictions in the PR body before running.
3. Minimal diff, one variable, tests. One branch per PR.
4. `/ci compare ppo --start-from h76h80yb --seeds 1 --total-timesteps 2000000`. Agent /ci commands are clamped (≤2M, ≤3 seeds; #372); owner-posted are not. Owner-scale compares outlive the report workflow (#384) — assemble the report manually from W&B.
5. Smoothed-final analysis (σ≈12, last-quartile) — never `summary="max"`.
6. Promising → 3-seed compare → paired stats → owner review. Refuted/neutral → post-mortem, close.
7. **Watch episodes before and after theorizing** (`scratchpad/watch_episodes.py`); qualitative diffs via `scratchpad/diff_checkpoints.py`; exact return math via `scratchpad/return_math.py` — **but only with env-stepped rewards, never a hand-rolled reward formula** (see metric semantics below).
8. Training is deterministic per (sha, seed). Pre-merge constants (STALE for new main): main@`cfc320e` 2M eval/thput 0.662/0.689/0.687; 10M lastQ 0.595/0.638/0.526. Checkpoints: `acc9wkc3` (main PPO@2M), `n7qggdmo` (KL-cap PPO@10M).

## METRIC SEMANTICS (hard-won 2026-08-10 — read before interpreting any plot)
- **`rollout/*` = temperature-1 SAMPLED episodes** (the training distribution, junk placements included). `rollout/*/reward` = mean undiscounted episode reward = terminal reward (reward is terminal-only). Verified by replication: 300 sampled CROSS episodes with `n7qggdmo` → mean 5.919 vs logged 5.92.
- **`eval/*` = greedy EOT-thresholded rollouts.** Different population; do not mix conclusions across the two.
[...]

PR #380, opened 2026-08-09
Vary the arm length in the MEMORISE lessons

beyarkay:
Closes part of #371.

Every `MEMORISE_N_INGREDIENT_RECIPES` arm used to be exactly
`marker → 1 belt → inserter → assembler` (and the mirror on the output side), so
the policy could learn the motif as a fixed stamp and reproduce it at every
marker on trial episodes. This widens the arm-length distribution so it can't.

## What changed

`build_memorise_recipes` (`factorion_rs/src/factory_gen.rs`) only — the recipe
pool, ingredient counts (one source per ingredient), one input inserter per arm,
one output inserter, lesson names and enum values are all untouched.

- Per arm, `belt_count` is drawn uniformly from `0..=MEMORISE_MAX_ARM_BELTS` (5),
  independently per arm.
- `belt_count == 0` puts the marker on the inserter's far cell: the input
  inserter picks up straight off the source, the output inserter drops straight
  onto the sink (sources and sinks connect like belts).
- `belt_count >= 1` places the marker at Manhattan distance `belt_count` from the
  far cell and fills the shortest walk between them — a straight line or a single
  L-bend, no staircases, detours or undergrounds. Belt directions and placement
  go through the existing `path_to_belts` / `place_belts`; the marker faces along
  the flow at its own end (source drops onto the arm, sink pulls off it).
- A draw that no marker cell can satisfy walks down to the first length that
  fits rather than rejecting the candidate (zero always fits — the far cell is
  the arm's own reservation), so crowded and tiny grids don't lose factories a
  shorter arm would have filled.
- Each arm's inserter and far cell are reserved before any routing, so a long arm
  can't take a cell a later arm has no alternative to. Collisions reject the
  candidate inside the existing rejection-sampling loop.
- The lesson now rejects on `unreachable != 0` as well as zero throughput, the
  same guard `CROSS_UNDER_BELT` and `FACTORY_1_INGREDIENT` use — longer arms give
  belts more ways to end up off a source→sink path, so the no-orphan invariant is
  enforced directly rather than implied by the geometry.

Per-arm belt counts come out essentially uniform on 0..5 (size 11, 150 seeds:
`MEMORISE_4` inputs 16.5 / 19.2 / 15.5 / 15.5 / 17.2 / 16.2 %), tilting slightly
short on tight grids where long arms get rejected more often.

Build rate is 200/200 seeds at sizes 7, 8, 11 and 15 for all four lessons, and
**improves** on main at smaller sizes (200 seeds each):

| size | kind | main | this PR |
|---|---|---|---|
| 5 | `MEMORISE_2` | 199/200 | 200/200 |
| 5 | `MEMORISE_3` | 82/200 | 200/200 |
| 5 | `MEMORISE_4` | 0/200 | 153/200 |
| 6 | `MEMORISE_4` | 69/200 | 196/200 |

A zero-belt arm needs two cells (inserter + far) where the old layout always
needed three (inserter + belt + marker), so the tight-grid cases that used to
fail now have somewhere to go.

## Tests

- Rust: `test_memorise_recipes_smoke` no longer asserts `belts == inserters`
  (still one inserter per arm, 9 assembler tiles, N sources, 1 sink). New
  `test_memorise_arms_vary_in_length_and_are_shortest_walks` walks each arm from
  its marker and asserts arms of 0 and >1 belts both occur across seeds, and that
  every arm is a straight or single-L walk.
- Python: `test_one_inserter_per_arm` bounds belts by
  `arms * MEMORISE_MAX_ARM_BELTS` instead of pinning them to `arms`.
  `test_arm_hops_account_for_every_belt` asserts the arms' graph hop counts sum
  to exactly the belts on the grid (so a bypassed arm or a stray belt fails), and
  `test_arm_lengths_vary_across_seeds` pins the observed length set to exactly
  `0..MEMORISE_MAX_ARM_BELTS` — which also fails loudly if the Rust cap drifts
  from the value mirrored in the test file.
- `test_unbuildable_kinds_dropped_not_hung` (`tests/test_sft.py`) used
  `MEMORISE_4` at size 5 as its can't-build example; that now builds, so it
  asserts on `SPLITTER_MERGE_SIDELOADED` and `FACTORY_1_INGREDIENT`, which never
  build there. The behaviour under test (drop, don't retry forever) is unchanged.
- `test_no_orphan_tiles_every_lesson` and the closed-form throughput test pass
  unchanged — the latter is what would catch cross-arm item mixing.

Checks run locally: `cargo fmt --check`, `cargo clippy -- -D warnings`,
`cargo test` (167 passed), `maturin develop --release`, `pytest tests/` (3272
passed, 708 skipped), `ruff check .`, `ty check .`. Ten factories per MEMORISE
kind were rendered and traced by hand: visibly different arm lengths, zero-belt
arms present, every arm a straight or single L, no stray tiles.

## Caveats for the owner

- **This changes the SFT data distribution.** Its real effect needs an SFT
  retrain — the current base checkpoint `h76h80yb` was trained on one-belt arms.
- **PPO compares against main are not clean on the `eval/MEMORISE_*` metrics**,
  because the eval factories themselves change; only same-side comparisons are
  meaningful there.
- **No `/ci` runs launched** — the SFT retrain is the owner's call.
- Worth a decision, not a bug: when every arm draws zero, the factory has no
  belts at all (~5% of `MEMORISE_1`, ~0.5% of `MEMORISE_2`, negligible above).
  Valid and it teaches direct marker↔inserter contact, but say the word if you'd
  rather guarantee at least one belt per factory.


beyarkay:
/ci compare sft --seeds 1

---

PR #383, opened 2026-08-09
Render the factories a compare's two sides actually build

beyarkay:
A compare report says whether `eval/thput` moved. It cannot say *which* factories moved or what the policy put on the grid. This adds the qualitative half.

## What's new

`factory_diff.py` — greedy-rollout N checkpoints per side over **one shared held-out factory set** (PPO's own eval set at a given seed), then render every factory the two sides consistently disagree on, side by side, largest gap first, under a per-lesson tally.

```
uv run python -m ci compare-renders <pr_ckpts> <main_ckpts> [--per-kind 50] [--out report.md]
```

Each spec is a local `.pt` path or a W&B run id; commas separate the seeds of one side. Works on any pair — SFT vs PPO, PPO@2M vs PPO@10M, seed vs seed.

## Multi-seed rule

A compare runs 3 seeds a side, so one seed landing elsewhere on one factory says nothing about the branch. A factory is reported **only when every run on one side beat every run on the other** (disjoint throughput ranges):

- main `0.5 0.5 0.5` vs PR `0.7 0.8 0.6` → reported
- main `0.5 0.5 0.5` vs PR `0.5 0.5 0.6` → not reported

Each block shows the median build per side with all its seeds' throughputs alongside; MAIN on the left, PR on the right, Δ = PR − main. Capped at 25 renders. With one seed a side this degenerates to "the throughput differs".

## CI wiring

`/ci compare` posts a second comment after the metric table. Ordering matters: the commit status is set **before** rendering, so a job timing out mid-render can never leave `factorion-ci/compare` pending; and the render sits before the assertion exit, so a red assert still gets its diff. A rendering failure is caught and logged — the metric report is never at risk.

Rendering needs torch + the Rust engine, installed on demand by `_ensure_render_deps()` in the dispatcher rather than as a workflow step: an `issue_comment` workflow always runs the **default branch's** YAML, so a step added on a PR branch would not exist until after merge, and the feature could never be tested on the PR introducing it. The probe runs in a child process — importing `factorion_rs` in the dispatcher binds the repo's own source directory (an empty namespace package) into `sys.modules`, where it shadows the wheel built moments later.

## Reuse

No new rollout loop: `sft.run_rollout_eval` gained an optional `records` list that collects each scored factory (seed, kind, thput, entity_cost, `render_factory` output) as its existing batched greedy rollout finishes a slot. `ci/report._fetch_group` was split so the run objects (not just their summaries) are reachable.

## Verification

Three `/ci compare sft --seeds 1 --num-samples 500000` runs on this PR; the first two found real bugs in the CI glue (missing workflow step, then the `sys.modules` shadowing), both fixed here. The third posts the diff correctly: 700 factories per checkpoint (50 × 14 lessons), 0 consistently different — the right answer, since this PR does not touch training and both sides' checkpoints are bit-identical.

Also verified locally: the diff on a real 3-seed PPO compare (`cmp-63cdc91-ppo-eb0a`, 6 checkpoints, 2 factories consistently different, both `CROSS_UNDER_BELT`), and the import behaviour in a bare torch-free venv. Full checklist green — ruff, ty, 3281 pytest, cargo fmt/clippy/test. Rollout cost measured at 75 s per checkpoint for 700 factories on 4 cores.

https://claude.ai/code/session_01AKgztKq3f4ZVRtqaPsWUJk


beyarkay:
/ci compare sft --seeds 1 --num-samples 500000

Smoke test for the factory-diff comment this PR adds: a short SFT compare against main, one seed a side. The metric table should be followed by a second comment with the side-by-side renders.

---

beyarkay:
/ci compare sft --seeds 1 --num-samples 500000

Re-run at `7a104ee`. The first attempt failed the render step with `ModuleNotFoundError: No module named 'torch'` (caught, so the metric report and commit status were unaffected): an `issue_comment` workflow always runs the **default branch's** YAML, so the install steps this PR added to `ci-command.yml` did not exist on the runner. Moved into the dispatcher, which does run from the PR head.

---

beyarkay:
/ci compare sft --seeds 1 --num-samples 500000

Third attempt, at `b34091c`. Attempt 2 installed torch and built the engine wheel successfully, then still died on `factorion_rs is out of date`: the dependency guard imported `factorion_rs` to decide whether to install, which bound the repo's own source directory (an empty namespace package) into `sys.modules` where it shadowed the wheel. The probe now runs in a child process. Verified locally in a bare torch-free venv: probe exits nonzero without torch, and once the wheel is installed the compiled module wins over the repo directory from the repo root.

---

PR #392, opened 2026-08-09
Show the policy and critic where items actually go

beyarkay:
Adds per-entity flow as a sixth observation channel, feeding the shared encoder so both the critic and the policy see it.

Related: **#391** takes the other projection of the same engine computation — entity flow as a terminal-reward term. We now share the same underlying semantics; see [the comment there](https://github.com/beyarkay/factorion/pull/391#issuecomment-5233897849) for the joint API shape and what each branch took from the other.

## What it is

The engine already propagates items/s through every graph node on its way to a sink — it just threw the intermediate rates away. This keeps them, paints each entity's total across its footprint, and returns the map from `simulate_throughput` alongside the score.

- **`FLOW` is an observation channel, not a world channel.** `Channel` stays at 5; the observation is `OBS_CHANNELS = 6`. Nothing places FLOW, it is stale the moment a tile changes, and `Factory.world_CWH` never carries it.
- **Encoded in eighths of an item/s**, clamped per entity to `MAX_ENTITY_FLOW = 30` (a saturated splitter). Integer-valued so SFT's uint8 demonstrations stay lossless, while still separating an inserter's 0.86 i/s from a belt's 15. `_encode_input` feeds it as `log1p(flow / FLOW_SCALE)`, a scalar plane alongside `FOOTPRINT`.
- **An entity carries items as soon as something feeds it** — reaching a sink is not required.
- **`factorion.observe(world_CWH) -> (obs, thput, num_unreachable)`** is now the single place the network's input is assembled — env, SFT demos, the builder UI, the mod server. It returns all three because the engine computes them in one pass, so nothing simulates twice.

No reward surface is touched. Everything here is input to the value function and the policy.

## Why partial credit, not just delivered flow

Probing a MOVE_ONE_ITEM build tile by tile (the markers are protected, so they are present throughout):

```
0/4 belts built   thput=0.0   flow=[30, 0,  0,  0,  0,  0]
1/4 belts built   thput=0.0   flow=[30, 15, 0,  0,  0,  0]
2/4 belts built   thput=0.0   flow=[30, 15, 15, 0,  0,  0]
3/4 belts built   thput=0.0   flow=[30, 15, 15, 15, 0,  0]
4/4 belts built   thput=15.0  flow=[30, 15, 15, 15, 15, 15]
```

Throughput is flat zero until the last placement; the channel advances a tile per belt. Requiring a source→sink path (the first version of this branch) collapsed rows 0–3 into identical zeros — silent for most of an episode, exactly where a value function most needs to separate "one belt short" from "nothing built". Nothing is lost by dropping it: delivery stays legible at the sink's own tile, so a chain broken at the last hop reads 15 along the belts and 0 at the sink.

## The case it catches

A belt loop scores zero throughput while `unreachable == 0` — every entity looks connected, and nothing in the old observation distinguishes it from a working line. The flow channel reads zero throughout. Covered by `test_flow_channel_zero_on_cycle` (Rust) and `test_flow_is_zero_on_a_belt_loop` (Python).

## Costs, measured

- **PPO rollout**: the engine now runs on every env step — the mid-episode skip is gone, since the agent's next input depends on the simulation the same way the terminal reward does. `env.step` with training diagnostics off: **36.6 → 77.6 µs/step**, i.e. **+0.17 s/iter** at 16 envs × 256 steps. Against the logged steady iteration (rollout ~1.36 s + update ~0.6 s) that is **~+8%** per iteration. The numpy diagnostics block stays skipped as before.
- **SFT demo generation**: **98 → 145 µs/pair** (10.2k → 6.9k pairs/s), in the DataLoader workers.

## Breaking: checkpoints

The encoder's `input_channels` goes 27 → 28, so **checkpoints from before this do not load**. `--start-from h76h80yb` fails with a shape mismatch, and so do `ci/sweep_ppo.yaml` and `scripts/factory_builder.py`'s default run id. PPO needs an SFT base retrained on this tip before it can be compared against main — worth doing anyway, since no run on record was trained at the current tip. #391 does not touch the network, so it can measure against `h76h80yb` today.

## Testing

New `tests/test_flow_channel.py` covers the engine → `observe` → env → encoder path: channel layout and dtype, flow following a belt line, an inserter bottleneck resolving to 0.875 i/s, an unfed entity at zero, the partial-chain progression above, a chain broken at the last hop, uint8 round-trip losslessness, the env's per-step refresh, the belt loop, and that changing flow alone moves both the encoder features and the critic's value. Six Rust unit tests cover `flow_channel` directly.

Full checklist green: `cargo fmt` / `clippy -D warnings` / `cargo test` (173), `maturin develop --release`, `pytest` (3296 passed, 708 skipped), `ruff`, `ty`.

https://claude.ai/code/session_0112XY1Ep6rtbzADJDE4HPRe


beyarkay:
/ci sft --num-samples 5000000

---

PR #394, opened 2026-08-11
Inserters pick up from both lanes, and from every item on a lane

beyarkay:
Inserters (regular and long-handed) now split their pickup rate evenly over both lanes of the belt they draw from, then split each lane's share across the items riding it in proportion to their rates. Drops are unchanged — still exactly one lane.

Found via `hcozpmwt`: two `TRIAL_RECIPE_TREE_DEPTH_1` builds grading `0.000` despite `reach 100%` — personal_roboport (seed 22) and pumpjack (seed 33).

## The bug, in two layers

Both builds wire four-ingredient recipes with only three inserters, so two ingredients share one belt: one feeds it **in-line from behind** (landing on both lanes), the other **sideloads from the side** (landing on one lane).

**Layer 1 — lane priority.** `pickup_input` gave the in-line pickup the belt's left lane first and greedily spent the whole 0.86 i/s there. The sideloaded ingredient never arrived, `min_ratio` collapsed to 0, and a fully-connected 22-entity factory crafted nothing.

**Layer 2 — item order within a lane.** Splitting evenly across lanes was not enough. The lane fed in-line carries *both* items, and its share still drained in item-id order, so whichever ingredient sorted first took the whole lane. Seed 33 lands exactly there: `steel_plate` (id 21) sorts before `pipe` (id 40) and both ride `b@5,2:R`, so pipe still arrived at zero and the pumpjack still scored `0.000`. Same starvation, one level down.

The priority rule mirrored real Factorio's near/left-lane preference, but as a steady-state flow model it is the wrong reading: it lets one lane — or one item — monopolise the hand forever. A real inserter's hand sweeps the whole tile and will not "pick up any items that cannot be inserted into the adjacent entity" ([wiki](https://wiki.factorio.com/Inserters)), so once the over-supplied ingredient's buffer saturates the swing goes to the other. One inserter feeding two ingredients off a mixed belt is a legitimate pattern; it costs throughput, it does not zero it.

## The change

`throughput.rs::pickup_input` allocates max-min fair across lanes, then proportionally across items within each lane:

- both lanes loaded, one item each → 0.43 each
- one lane loaded → that lane covers the full 0.86 (unchanged)
- both lanes carrying the same item → 0.43 + 0.43 = 0.86 (unchanged)
- one lane mixed 50/50 → 0.215 of each from that lane

Only **mixed-item** pickups move. `GraphNode::direction` and `curved` existed solely to resolve the old priority and are removed with it.

| build | before | after | binds on |
|---|---|---|---|
| seed 22 personal_roboport | 0.0 | 0.005375 | gears at 0.215/40 |
| seed 33 pumpjack | 0.0 | 0.0215 | pipe at 0.215/10 |

In both, the in-line ingredient occupies both lanes and collects 0.645 while the sideloaded one collects 0.215 from its single lane — so the sideloaded ingredient binds. Four dedicated inserters would reach 0.0191 and 0.086 respectively; sharing an inserter costs real throughput, which is the point.

## Blast radius: none on training

Compared lesson generation across **all 14 `LessonKind`s × 40 seeds** (560 samples), before vs after, re-run after each layer:

```
0 of 560 lesson samples changed
```

Identical layouts, identical `max_throughput`, identical entity counts. Lesson belts carry a single item per lane, so both splits total exactly what priority did. SFT/PPO normalization denominators and the data distribution are untouched — no baseline shifts, `h76h80yb` stays comparable.

## Tests

- `test_pickup_splits_evenly_across_lanes` — in-line and perpendicular inserters both read 0.43 of each item off a two-item belt (replaces the two lane-priority tests, whose premise is gone).
- `test_pickup_falls_back_to_other_lane` — unchanged, now covering the max-min fallback.
- `lane_isolated_pickup.yaml` → `lane_pickup_split.yaml` — the fixture asserted that two inserters could *unmix* a two-item belt by pickup priority. They no longer can; it now asserts both get the same 50/50 mix at 0.43.
- `mixed_belt_pickup.yaml` (new) — the seed-22 and seed-33 grids as built, plus a stripped-down electronic_circuit version pinning the same arithmetic without the 11×11.

Docs updated to match: the `> **In Factorion:**` notes in `wiki/inserter.md`, `wiki/transport-belt.md`, `wiki/two-lane-system.md`, and the dual-lane bullet in `CLAUDE.md`.

## Known limitation, not addressed here

A lane node caps flow **per item**, not in total, so `b@5,2:R` reports steel_plate at 7.5 *and* pipe at 7.5 — 15 i/s on a lane whose physical cap is 7.5. Proportional pickup reads the right *ratio* off that regardless, so this PR is correct either way, but capping a lane's total would also give sideloading-onto-a-saturated-lane its real behaviour. Worth its own change.

Separately, trial `max_throughput` is the recipe's nominal rate (1.0 i/s), which for personal_roboport would need 45 battery/s through 0.86 i/s inserters — ~53 inserters of battery alone. A flawless seed-22 build measures 0.0191, so it displays as `0.02 of 1.00`; every trial score is compressed into the bottom couple of percent before build skill enters the picture.

## Checklist

`cargo fmt` · `cargo clippy -D warnings` · `cargo test` (167 passed) · `maturin develop --release` · `pytest tests/` (3288 passed, 708 skipped) · `ruff check .` · `ty check .` — all pass. All three commits pass standalone.

(`cargo clippy --all-targets` reports 2 `iter_cloned_collect` lints in `types.rs` test code; both pre-date this branch — 3 before, 2 after — and are outside the checklist's lib-only command.)


beyarkay:
/ci compare sft --seeds 1 --num-samples 2500000

